PR #3, opened 2025-05-21
[Chore] WORKATY-61: Import Part 1 components into Chromatic

Alimnfll:


adamnoto (changes requested):
Great! Just a little request. Let's please:

1. Rename CapsuleTab -> Capsule; we'll have CapsuleTab later on.

2. Rename ProfileFrame -> Avatar, much more concise. Then let's add props `size` which is optional, to Avatar, so we can specify the size in pixel. If not given, let's render the current size as-is.

Alimnfll:
> Great! Just a little request. Let's please:
> 
> 1. Rename CapsuleTab -> Capsule; we'll have CapsuleTab later on.
> 2. Rename ProfileFrame -> Avatar, much more concise. Then let's add props `size` which is optional, to Avatar, so we can specify the size in pixel. If not given, let's render the current size as-is.

ofc, done

adamnoto (approved):

PR #4, opened 2025-05-23
[Chore] WORKATY-52: Develop Our Design Language on Story Book: Part 2

Alimnfll:


adamnoto (changes requested):
Thank you for the work. Have some requests:

1. When hovering over "dangerous" tooltip menu item, can we please blend the border to the same red color?

    Screenshot 2025-05-24 at 7 46 04

2.  When selecting all items, the checkbox on the header should be checked in full, instead of partial:

    Screenshot 2025-05-24 at 7 49 03

    And then, when only partial items are checked, the checkbox on the header should checked partially, instead of in full:

    Screenshot 2025-05-24 at 7 49 13

    Apologize that the design spec was incomplete/incorrect in showing the checkbox state at the header. I just realized that 😅 

3. Can we please change the tooltip menu look and feel to be similar of the Workaty's menu, the one at the header? Similar in terms of appearance.

4. Let's apply `w-full` automatically on `Toolbar` so we don't have to specify it
5. Let's introduce `variant` props to `Toolbar.Menu`, which is optional and default to `default`. Another variant is `primary` which has background `RGB(86, 154, 236)` and when on hover: `RGB(64, 142, 232)`; and also `danger`: `RGB(206, 55, 93)` background, on hover: `RGB(200, 53, 50)` background. Add a new example of toolbar menu with variant.
6. Can we please style toolbar to have similar look and feel as per the specification. Let's have a little bit of visual cue, so let's apply some shadow and so on. Let's use border-radius of 2px.
7. Let's pad the table rows with 3px instead of 4px, 4px seems a little bit too sparse.
8. Can we try using alternate color for table rows? A light gray should work for all even rows, so we can use `table tr:nth-child(even)` or something like that to target those even rows and apply specific background color.

Thanks :)

adamnoto:
Hmm, for point 6:

> Can we please style toolbar to have similar look and feel as per the specification. Let's have a little bit of visual cue, so let's apply some shadow and so on. Let's use border-radius of 2px.

After second thought, let's introduce `<Card>` component, which render a `<div>` container with shadow, inset or styling necessary to render similar to the document, leaving the `Toolbar` themselves plain as it is now. We can then achieve this styling in a more modular manner:

```tsx
<Card shadow="xs">
  <Toolbar>
    /* wrap tooltip menu inside a Card, where the shadow is this cardShadow */
    <ToolbarMenu ... submenuCardShadow="sm" />
  </Toolbar>
</Card>
```

The default of `shadow` in `<Card>` is `sm`, and the default `submenuCardShadow` for `<ToolbarMenu>` is `"none"`. The possible shadow for card is: `none`, `xs`, `sm`, `md`, and `lg`. When `none`, a `Card` should just render its children, without any wrapping, only when it's not `none` that a card should render itself and its children.

image

The above is `xs`-level shadow. `sm`-level shadow should be ~20% more shadow, and `md`-level shadow is 50% shadow and `lg` should be 70% more shadow than `xs`.

So, point 3 and point 6 in the comment above is not needed. Let's just introduce `Card`.

adamnoto (changes requested):
Nice, just a little bit more request.

1. Can we please render the chevron color to white when on primary?
    
    Screenshot 2025-05-26 at 16 52 50

    Also let's add vertical bar before the chevron, like this:

    Screenshot 2025-05-26 at 16 59 47


2. On hover, can we just highlight background to soft-gray `#f2f2f2` on default? But on danger, with `#e71f29` ? The black border is giving a great contrast, stylistically.

3. Let's rename the story from Tooltip Menu to TipMenu ?

4. On `ToolbarMenu` can we remove the shadow styling? The shadow styling should be just provided by `<Card>`

    Screenshot 2025-05-26 at 16 59 15

PR #8, opened 2025-05-31
[Chore] WORKATY-53: Story Book: Part 3

Alimnfll:


adamnoto (changes requested):
Other than these, so far so good functionality-wise 💯 will review on the code side later.

1. When clicking on `Today` on Calendar, the today's date should be circled.
2. When hovering on Calendar's month-year label:
  
    Screen Shot 2025-06-03 at 7 40 57 PM

    Let's render some grayish border color, then, when it's clicked, let's render something like this:

    Screen Shot 2025-06-03 at 7 42 05 PM

3. The behavior above should be the same with the Datebox.
4. For `LoadingSpinner` storybook, could you make the `label` editable? so we can type and see how it would render with a label? And maybe let's also add an example with label :)
5. For the `TreeList` could we please render something similar to the example in linear? When we hover, let's add the small side blue bar, and also when the item is active/selected. Maybe we need to introduce `isSelected` on each item, and if it's true, we render the blue bar.
6. For capsule tab, can we add a little padding, and a horizontal line divider between the tab and the content?

    Screen Shot 2025-06-04 at 6 58 47 AM
7. For `FileInputBox`, let's use the same border style as the drop box?

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+  disableWeekend,
+  minDateOffset = 80,
+  maxDateOffset = 50,
+  format = "mm/dd/yyyy",
+  className,
+}: CalendarDrawerProps) {
+  const STATE_DATE = inputValue !== "" ? new Date(inputValue) : new Date();
+  const [currentDate, setCurrentDate] = useState(STATE_DATE);
+
+  const monthCurrent = STATE_DATE.toLocaleString("default", {
```
adamnoto, line 72: `currentMonth`

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+    month: "short",
+  });
+  const yearCurrent = STATE_DATE.toLocaleString("default", {
```
adamnoto, line 75: `currentYear`

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+  disableWeekend?: boolean;
+  minDateOffset?: number;
+  maxDateOffset?: number;
+  format?: FormatProps;
+  className?: string;
+}
+
+interface OpenBoxProps {
+  open: boolean;
+  month: string;
```
adamnoto, line 44: can we use integer for the month? with `0` internally indicate `January`, and `11` for `December` so that we don't have to do `findIndex` on an array

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+        const day = Math.min(
+          currentDate.getDate(),
+          new Date(yearNumber, month + 1, 0).getDate()
+        );
+        setCurrentDate(new Date(yearNumber, month, day));
+      }
+    }
+  };
+
+  const getDaysInMonth = (date: Date) => {
```
adamnoto, line 121: `getDatesInMonth`

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+    const year = date.getFullYear();
+    const month = date.getMonth();
+    const days: Date[] = [];
```
adamnoto, line 124: `dates`

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+    const lastDay = new Date(year, month + 1, 0).getDate();
+    for (let i = 1; i <= lastDay; i++) {
+      days.push(new Date(year, month, i));
+    }
+    return days;
+  };
+
+  const days = getDaysInMonth(currentDate);
```
adamnoto, line 132: `dates`

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+  getFloatingProps?: (
+    userProps?: HTMLAttributes<HTMLUListElement>
+  ) => HTMLAttributes<HTMLUListElement>;
+  refs?: { setFloating: Ref<HTMLUListElement> };
+  floatingStyles?: CSSProperties;
+  listRef?: MutableRefObject<(HTMLLIElement | null)[]>;
+  dayNames: string[];
+  monthNames: string[];
+  disableWeekend?: boolean;
+  minDateOffset?: number;
```
adamnoto, line 36: `yearPastReach`

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+  maxDateOffset?: number;
```
adamnoto, line 37: `yearFutureReach`

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+    setInputValue(formatDate(today, format));
+    setHighlightedIndexChange(0);
+    setIsBoxOpen((prev) => ({
+      ...prev,
+      month: monthCurrent.toUpperCase(),
+      year: yearCurrent,
+    }));
+  };
+
+  const prevClickDate = new Date(
```
adamnoto, line 162: `prevMonth`

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+    currentDate.getFullYear(),
+    currentDate.getMonth() - 1,
+    1
+  );
+
+  const nextClickDate = new Date(
```
adamnoto, line 168: `nextMonth`

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+                  year: "numeric",
+                })
+                .toUpperCase()}
+            </div>
+          ) : (
+            <div className="flex flex-row gap-1 w-full">
+              <Combobox
+                options={monthNames}
+                inputValue={isBoxOpen.month}
+                placeholder="JAN"
```
adamnoto, line 295: set to first element of the month names array

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+                setInputValue={(value) => {
+                  onChangeValueDate({
+                    target: { name: "month", value },
+                  } as ChangeEvent<HTMLInputElement>);
+                }}
+              />
+              <Combobox
+                options={yearOptions}
+                inputValue={isBoxOpen.year}
+                placeholder="2025"
```
adamnoto, line 306: the year of the `currentDate`

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+interface OpenBoxProps {
+  open: boolean;
+  month: string;
+  year: string;
+}
+
+type FormatProps = "mm/dd/yyyy" | "yyyy-mm-dd" | "dd/mm/yyyy";
+type DateBoxOpen = "open" | "month" | "year";
+
+function CalendarDrawer({
```
adamnoto, line 51: let's call this `Calendar` and the props is `CalendarProps` ... it is not a drawer in this context

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+    case "dd/mm/yyyy":
+      return `${day}/${month}/${year}`;
+    case "mm/dd/yyyy":
+    default:
+      return `${month}/${day}/${year}`;
+  }
+}
+
+const Calendar = CalendarDrawer;
+export { Calendar, CalendarDrawer };
```
adamnoto, line 469: let's not export CalendarDrawer here, since it's not a drawer; that other components think of it as a drawer is not the business of this component :)

components/combobox.tsx
```diff
@@ -1,214 +1,109 @@
...
-  RiArrowDownSLine,
-  RiArrowUpSLine,
-  RiCloseLine,
-} from "@remixicon/react";
+import { CSSProperties, HTMLAttributes, MutableRefObject, Ref } from "react";
+import { Selectbox } from "./selectbox";
 
 interface ComboboxProps {
   options: string[];
+  type?: "default" | "calendarItems";
```
adamnoto, line 6: when is this `"calendarItems"` used? in general, the combobox shouldn't even care/know if it is calendar items or etc...

components/filedropbox.tsx
```diff
@@ -0,0 +1,226 @@
...
+  handleOnComplete?: (props: HandleOnCompleteProps) => void;
+  setProgressPercentage?: (props: number) => void;
+  progressPercentage?: number;
+}
+
+export default function FileDropBox({
+  classNameContainer,
+  placeholder = "Drag and Drop Your File",
+  accept = "*",
+  handleFileDropped,
```
adamnoto, line 34: `onFileDropped`

components/filedropbox.tsx
```diff
@@ -0,0 +1,226 @@
...
+  handleOnComplete,
```
adamnoto, line 35: `onComplete`

components/filedropbox.tsx
```diff
@@ -0,0 +1,226 @@
+import { RiFile2Line, RiFileUploadLine, RiImageLine } from "@remixicon/react";
+import { cn } from "./../lib/utils";
+import { ChangeEvent, DragEvent, useRef, useState } from "react";
+import LoadingSpinner from "./loading-spinner";
+
+export interface HandleFileDroppedProps {
```
adamnoto, line 6: `OnFileDroppedFunctionProps`

components/filedropbox.tsx
```diff
@@ -0,0 +1,226 @@
...
+  files: File[];
+  succeed: (file: File) => void;
+  error: (file: File, errorMessage: string) => void;
+  setProgressLabel: (label: string) => void;
+  progressPercentage?: number;
+}
+
+export interface HandleOnCompleteProps {
```
adamnoto, line 14: `OnCompleteFunctionProps`

components/filedropbox.tsx
```diff
@@ -0,0 +1,226 @@
...
+      size: 50,
+    },
+  ];
+
+  const fileInputRef = useRef<HTMLInputElement>(null);
+  const [isDragging, setIsDragging] = useState(false);
+  const [currentIndex, setCurrentIndex] = useState<number | null>(null);
+  const [progressComponentLabel, setProgressComponentLabel] = useState("");
+  const [progressPercentage, setProgressPercentage] = useState(0);
+  const [progress, setProgress] = useState<"idle" | "loading" | "succeed">(
```
adamnoto, line 60: let's constantize / typify the idle, loading, succeed, and use the type here instead of directly `useState<a|b|c>`

components/filedropbox.tsx
```diff
@@ -0,0 +1,226 @@
...
+    "idle"
+  );
+  const [errorMessage, setErrorMessage] = useState<string[]>([]);
```
adamnoto, line 63: `errorMessages`

components/filedropbox.tsx
```diff
@@ -0,0 +1,226 @@
...
+    };
+
+    const succeed = (file: File) => {
+      succeedFiles.push(file);
+      handleUpdateProgressBar(succeedFiles.length, total);
+    };
+
+    const error = (file: File, errorMessage: string) => {
+      failedFiles.push(file);
+      console.error(errorMessage);
```
adamnoto, line 118: no longer needed?

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+  useEffect,
+  Ref,
+  ChangeEvent,
+} from "react";
+import { Button } from "./button";
+import { cn } from "../lib/utils";
+
+import Combobox from "./combobox";
+
+interface CalendarDrawerProps {
```
adamnoto, line 20: a lot of the props here inherits from the `children` prop of the `Selectbox` I believe... can we refactor? ie:

```tsx
interface SelectboxChildrenProps {
  highlightedIndex?: number
  ...
}

interface CalendarDrawerProps extends SelectboxChildrenProps {
  ...
}
```

and for Combobox and any other things were props are inherited for selectbox.

components/treelist.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+type Story = StoryObj<typeof TreeList>;
+
+export const Default: Story = {
+  render: () => {
+    const setPerson = (item: { id: number; label: string }) => {
+      console.log("Clicked person:", item.label);
+    };
+
+    const DATA_TREE_LIST = [
```
adamnoto, line 20: `TREE_LIST_DATA`

components/treelist.tsx
```diff
@@ -0,0 +1,98 @@
+import { ReactNode, useState } from "react";
+import { cn } from "./../lib/utils";
+
+interface TreeListProps<T extends TreeListItemsProps> {
+  content: Array<{
+    label: string;
+    items: T[];
+  }>;
+  classNameContainer?: string;
```
adamnoto, line 9: `containerClassName`

adamnoto (changes requested):
Good job. Just some small behavior request, and code changes 😺 

1. We need to improve the way Combobox works. When we reveal the drawer for the first time, even when there's already text there, let's show all the possible options, rather than filtered out, with the selected item having a firmer shade of blue (#61A9F9) than the hover/the rest. Only if the user type any word should we do the filtering.

    So, when opening a combobox, instead of showing this:

    Screen Shot 2025-06-05 at 4 44 17 AM

    Always show all the possible choices, with selected one rendered in #61A9F9. Only filter when the user change/type some value into the combobox. This is because, so that user can change selected option easily, without having to clear out the value. If the user wanted to narrow down the shown values, they can start typing.

2. Let's also add more test :)
3. Can we make it so that text selection is disabled on tree list, button, capsule, tip menu, and calendar? In the future, let's generally disable text selection, except for most of content/input elements.
4. Can we make new control values editable for new components?

    Screen Shot 2025-06-05 at 4 38 54 AM

    As we can see above, the control values are disabled. Let's enable them so that we can make some testing immediately on the storybook.

5. The `<` and `>` must always work regardless if they're out-of-bound from the year list or not, so let's say 1945 is the last selectable year, the user can still click < in the month of January 1945, and they'll arrive at December 1944.

    Screen Shot 2025-06-05 at 5 16 37 AM

    The `<` above shouldn't be disabled.
6. Rename file to be consistent:
    - `components/filedropbox.tsx` -> `components/file-drop-box.tsx`
    - `components/fileinputbox.tsx` -> `components/file-input-box.tsx`

components/filedropbox.tsx
Alimnfll: nah, it's just for now, it's needed bcs we can detect and make add condition for error on later

components/calendar.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+    const leftChevron = await canvas.findByLabelText("Previous Month");
+    await userEvent.click(leftChevron);
+
+    const dayButton = await canvas.findByText("13");
+    await userEvent.click(dayButton);
+  },
+};
+export const CalendarWithDisableWeekend: Story = {
+  render: () => {
+    const DAYS_NAME_ITEMS = ["Su", "Mo", "Tu", "We", "Th", "Fr", "Sa"];
```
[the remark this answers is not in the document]
Alimnfll: done

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+  setInputValue: (value: string) => void;
+  inputValue: string;
+  setIsOpen?: (open: boolean) => void;
+  getFloatingProps?: (
+    userProps?: HTMLAttributes<HTMLUListElement>
+  ) => HTMLAttributes<HTMLUListElement>;
+  refs?: { setFloating: Ref<HTMLUListElement> };
+  floatingStyles?: CSSProperties;
+  listRef?: MutableRefObject<(HTMLLIElement | null)[]>;
+  dayNames: string[];
```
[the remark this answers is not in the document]
Alimnfll: done

components/calendar.tsx
```diff
@@ -0,0 +1,469 @@
...
+  monthNames: string[];
```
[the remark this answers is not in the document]
Alimnfll: done

components/calendar.tsx
Alimnfll: done

components/calendar.tsx
Alimnfll: done

components/calendar.tsx
Alimnfll: i think for month why we should use by string? that's because on combobox just accepted format string, but i'll try it with interger format.

components/calendar.tsx
Alimnfll: done

components/calendar.tsx
Alimnfll: done

components/filedropbox.tsx
Alimnfll: done

components/filedropbox.tsx
Alimnfll: done

components/filedropbox.tsx
Alimnfll: done

components/treelist.stories.tsx
Alimnfll: done

components/calendar.tsx
Alimnfll: done

components/calendar.tsx
Alimnfll: yaps, that's always currentDate, because we write on useState isBoxOpen.year state.

components/calendar.tsx
Alimnfll: done

components/calendar.tsx
Alimnfll: done

adamnoto (approved):
It's lit 🔥 just some small concerns that could be addressed in a future revision or as part of part 4 or 5; they're just small reorganization.

components/calendar.tsx
```diff
@@ -0,0 +1,486 @@
...
+  RiCheckLine,
+} from "@remixicon/react";
+import { useState, useEffect } from "react";
+import { Button } from "./button";
+import { cn } from "../lib/utils";
+import Combobox from "./combobox";
+import { OptionsProps } from "./selectbox";
+import { CalendarDrawerProps } from "./datebox";
+
+export interface CalendarProps extends CalendarDrawerProps {
```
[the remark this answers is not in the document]
Alimnfll: i think was done

components/selectbox.tsx
```diff
@@ -0,0 +1,258 @@
...
+  RiCloseLine,
+} from "@remixicon/react";
+import { cn } from "./../lib/utils";
+
+export interface OptionsProps {
+  text: string;
+  value?: string | number;
+}
+
+export interface FloatingUIProps {
```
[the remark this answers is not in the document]
Alimnfll: ofc, done

PR #11, opened 2025-07-02
[Feature] WORKATY-68: Story Book: Part 6

Alimnfll:


components/error-slate.tsx
```diff
@@ -0,0 +1,32 @@
+import { ReactNode } from "react";
+
+interface ErrorSlateProps {
+  code?: "404" | "500" | "403";
```
adamnoto, line 4: let's have all full error codes supported (including `400`, `401`, `402`, etc).

components/togglebox.tsx
```diff
@@ -0,0 +1,73 @@
...
+  onChange?: (e: ChangeEvent<HTMLInputElement>) => void;
+  icon?: RemixiconComponentType;
+  isLoading?: boolean;
+  name?: string;
+}
+
+export default function Togglebox({
+  checked = false,
+  onChange,
+  icon: Icon = RiLock2Fill,
```
adamnoto, line 18: let's make it optional, default to none (not rendering any icon)

components/togglebox.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { RemixiconComponentType, RiLock2Fill } from "@remixicon/react";
+import { cn } from "./../lib/utils";
+import LoadingSpinner from "./loading-spinner";
+
+interface ToggleboxProps {
+  checked?: boolean;
+  onChange?: (e: ChangeEvent<HTMLInputElement>) => void;
+  icon?: RemixiconComponentType;
+  isLoading?: boolean;
+  name?: string;
```
adamnoto, line 12: let's add `id` as optional too since we can pass in `id` to `input`

components/togglebox.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export default function Togglebox({
+  checked = false,
+  onChange,
+  icon: Icon = RiLock2Fill,
+  isLoading = false,
+  name = "togglebox",
```
adamnoto, line 20: is it possible not to set default `name` and `id` if not provided?

adamnoto (changes requested):
Thanks, looking awesome! A bit comment on the code, and UX interactivity.

1. for `Togglebox`, the `icon` should be optional, and blank by default, and when not provided, should not render any icon. also let's add `id` for the input ID, and `label` for rendering text/caption next to togglebox.
2. for list, when `openerStyle` is set to `togglebox` , let's render it without icon (so no lock)
3. great `ErrorSlate`, but could we customize the color, maybe let's expose cubeFaceStyle which allow user to supply css? and, let's default the style to background red #dd0b0b, border-width 1px and border-color #a80000 

    Screen Shot 2025-07-02 at 7 57 32 PM

4. For the `<Timeline>` example, can we have the first 3 `completed` ... so the last one is not completed. If we hover on the `<Timeline.Item>` let's have "outer circle" with background `#4eb59c`, with enlarge animation, if `completed`, otherwise if not completed, when hovered, the outer circle is black, with enlarge animation too on hover. let's change the completed background color of the circle and the line to `#17a114`. if it is hard to understand, please take a look at this video:


    https://github.com/user-attachments/assets/84318aa1-5503-4b30-90f1-2d4238214741

    let's make it consitent for the `Stepline` too, with hover and everything as we discussed above:

    Screen Shot 2025-07-03 at 7 10 14 AM

5. For `<Separator>` let's have `textStyle` with which we can provide styling. Also, let's have `variant` that can be "xs" "sm" "md" "lg" for the font size, the default one (the current one) should be just "sm"

6. Let's move the Crumb story to Control, and name it "Crumb" not "Crumbs" :D can we when clicking the "..." expand it with smooth animation?

Yep's that's all for now :)

components/error-slate.tsx
Alimnfll: done

components/togglebox.tsx
Alimnfll: done

adamnoto (changes requested):
Looks good. Just a bit more. 🙏 

1. let's make the size of the `Timebox` like that of `Textbox`, currently it's too big... also when on focus, no need to render yellowish background, and please render the same blue border like in other input box when in focus.

    I have an idea, is it possible that: if we click on the hour/minute/second, the value doesn't automatically disappear? let's make them disappear when we type the first character. so say, the value is currently 12, as soon as I type 2, then the whole value for that part is 2, if I type 3, then becomes 23.

2. can the toggle, when activated, a bit more to the right so that there's no blue-ish remnant rendered? currently it looks a bit off when activated.

    Screen Shot 2025-07-03 at 10 46 12 PM

    Maybe, just maybe, make the circle bigger, so that it's engulfing the control in the background, ensuring there's no color leakage. Perhaps.

1. For the last `<Timeline.Item>` please don't extend the line beyond the content. Currently, the line looks way too long while there's not next item/content; it looks hanging.

    Screen Shot 2025-07-03 at 10 49 43 PM

1. Sorry for this, but can we remove `completed` prop for `Stepline` and `Timeline` BUT we will have `variant`, which can be: `completed`, `current`, `todo`, and `error`. 

    - For `current`, let's render the background and the connecting line to `#00b62e` and outer circle to `#2fe620` on hover.
    - For `error`, let's render the background to and the connecting line to `#b60000` and outer circle to `#8f0751`
    - For `completed` same with current `completed=true` and for todo same with `completed=false` currently

    This is a much reliable system, so that we can add different coloring for warning in the future, but for now this is enough.

Overall, it's lit 🔥

components/timebox.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+  render: (args) => {
+    const [value, setValue] = useState({ timebox: "" });
+
+    const onChangeValue = (e: ChangeEvent<HTMLInputElement>) => {
+      const { name, value } = e.target;
+
+      setValue((prev) => ({ ...prev, [name]: value }));
+    };
+
+    console.log(value);
```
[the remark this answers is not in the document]
Alimnfl: done

components/timebox.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
```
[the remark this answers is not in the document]
Alimnfl: done

components/error-slate.tsx
```diff
@@ -0,0 +1,89 @@
...
+    | "505";
+  children?: ReactNode;
+  title?: string;
+  cubeFaceStyle?: CSSProperties;
+}
+
+function ErrorSlate({
+  code,
+  children,
+  title = "PageNotFound",
```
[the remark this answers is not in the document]
Alimnfl: done

components/timebox.tsx
```diff
@@ -0,0 +1,180 @@
...
+import { cn } from "./../lib/utils";
+import rawDayjs from "dayjs";
+import utc from "dayjs/plugin/utc";
+import timezone from "dayjs/plugin/timezone";
+
+rawDayjs.extend(utc);
+rawDayjs.extend(timezone);
+const dayjs = rawDayjs;
+
+interface TimeboxProps {
```
adamnoto, line 11: can we please accept `value` in the format of `hh:mm:ss` so if `value="12:25:30"` it correctly can set the hour minute and second components accordingly

components/togglebox.tsx
```diff
@@ -0,0 +1,99 @@
...
+  titleClassName,
+  ...props
+}: ToggleboxProps) {
+  const inputId = `togglebox-${name}-${props.value}`;
+
+  return (
+    <div className={cn("flex flex-row gap-2", containerClassName)}>
+      <AnimatePresence mode="wait" initial={false}>
+        {isLoading ? (
+          <motion.div
```
adamnoto, line 37: now I wonder, when `isLoading`, I think to avoid flickering, how if we render the `LoadingSpinner` on the togglebox circle slider?

to demonstrate this, lets update the isloading story too, and sleep 1 second before setting the isloading to false, when sliding it...

components/stepline.tsx
```diff
@@ -0,0 +1,152 @@
...
+            variant === "error" && "bg-[#b60000]",
+            (variant === "completed" || variant === "current") && "bg-[#00b62e]"
+          )}
+        >
+          {id}
+        </div>
+      </div>
+      {(title || subtitle) && (
+        <div className="flex flex-col">
+          {title && <h2 className="font-medium text-sm">{title}</h2>}
```
adamnoto, line 137: let's use `<span>` instead of `h2` ... it's ok, let's use `h2` only for actual page section :)

components/stepline.tsx
```diff
@@ -0,0 +1,152 @@
...
+export interface SteplineProps {
+  children?: ReactNode;
+  className?: string;
+  reversable?: boolean;
+}
+
+export interface SteplineItemProps {
+  title?: string;
+  subtitle?: ReactNode[];
+  variant?: "current" | "todo" | "error" | "completed";
```
adamnoto, line 20: can we extract it to exportable `SteplineVariant` type, and then use the type on `Timeline` ? to ensure consistency.

components/stepline.tsx
```diff
@@ -0,0 +1,152 @@
...
+                hoveredIndex: hoveredIndex,
+              })}
+            </div>
+            <div className="flex flex-row relative items-center w-full">
+              {index !== childArray.length - 1 && (
+                <div
+                  className={cn(
+                    "flex-1 h-px min-w-[44px] bg-gray-400",
+                    variant === "error" && "bg-[#b60000]",
+                    (variant === "completed" || variant === "current") &&
```
adamnoto, line 82: is it possible to have function like `getBackgroundColorByVariant` and `getInnerCircleBackgroundColorByVariant` and use those functions on stepline and timeline to ensure coloring consistency.

if it's possible doing so, but by not using `bg-[]` style of setting, I don't mind specifying the coloring as `style={{}}` kind of setting.

components/timebox.tsx
```diff
@@ -0,0 +1,180 @@
...
+      if (type === "hour") setHour("");
+      if (type === "minute") setMinute("");
+      if (type === "second") setSecond("");
+      return;
+    }
+
+    if (newDigit === "" || isNaN(Number(newDigit))) return;
+
+    let newVal = "";
+    let max = 0;
```
adamnoto, line 64: I think, we don't need to variable, let's define const like `MAX_HOUR_VALUE`, `MAX_SECOND_VALUE` and `MAX_MINUTE_VALUE` and use them instead

components/togglebox.tsx
```diff
@@ -0,0 +1,99 @@
...
+  onChange,
+  icon: Icon = null,
+  isLoading = false,
+  label,
+  description,
+  containerClassName,
+  titleClassName,
+  ...props
+}: ToggleboxProps) {
+  const inputId = `togglebox-${name}-${props.value}`;
```
adamnoto, line 31: let's not do this, we can rather accept an optional `id` props, and hand them down to the `input` component correctly, let's trust our user they can set the `id` and we should just pass them as-is without modifying

components/stepline.tsx
```diff
@@ -0,0 +1,152 @@
...
+  title,
+  variant = "todo",
+  className,
+  id,
+  hoveredIndex,
+}: SteplineItemProps &
+  Partial<{
+    hoveredIndex?: number | null;
+  }>) {
+  const steplineItemId = `stepline-item-${id}`;
```
adamnoto, line 106: let's pass the `id` as-is :)

components/timeline.tsx
```diff
@@ -0,0 +1,162 @@
...
+  title,
+  sidenote,
+  className,
+  id,
+  onClick,
+  isClickable,
+}: TimelineItemProps & {
+  isClickable?: boolean;
+}) {
+  const timelineItemId = `timeline-${id}`;
```
adamnoto, line 127: let's pass the `id` as-is

adamnoto (changes requested):
Aren't they absolutely supeerrb? Just have only a bit request please:

1. On `ErrorSlate` story, could we make the story editable/experimentable, let's use code `404` on the story, and set the title to `PageNotFound`.

1. On `Stepline` and `Timeline` let's update inner circle background for `error` variant to `#ff0000` ? I think this red is brighter. Sorry :)

    
    Screen Shot 2025-07-05 at 12 27 44 AM
    Screen Shot 2025-07-05 at 12 15 04 AM

1. On `Timebox` apart from the code review below, could we please make the `:` separator be on the same level, or a little bit up... currently it's very nestled down.

    Screen Shot 2025-07-04 at 11 58 36 PM

    Also, could we not reset the value just by clicking it/changing the focus on any time component? Let's reset the value on the first valid keydown/keypress (so pressing character or space shouldn't change the value), and on the second keypress append the value as long as it's a valid unit of time for that time component (ie 23 max for hour, 59 max for minute and second, etc kind of rule)

    Also, I am wondering if it is possible or this is a browser behavior. If possible, when hovering, and the value is empty, let's no need to change the text color on the `HH` or `MM` or `SS` wording. That is, when it's empty, let's render the placeholder with consistent color as for when hovering. Please use the lighter gray color we currently used on hovering as the consistent placeholder color in this case (if possible).

components/timebox.tsx
```diff
@@ -0,0 +1,180 @@
...
+    onChange?.(valueTime);
+  };
+
+  const inputClass = cn(
+    "w-[50px] text-center border-none items-center px-1 py-1 text-sm bg-white border border-gray-300 focus:outline-none placeholder:text-center",
+    "appearance-none [-moz-appearance:textfield] [&::-webkit-inner-spin-button]:appearance-none [&::-webkit-outer-spin-button]:appearance-none",
+    editable && "hover:text-gray-500"
+  );
+
+  console.log(hour);
```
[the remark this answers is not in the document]
Alimnfl: done

components/stepline.tsx
Alimnfl: done

components/timebox.tsx
Alimnfl: done

components/timebox.tsx
Alimnfl: done

components/stepline.tsx
Alimnfl: this is still on stepline component?

components/timebox.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+      const interval = setInterval(() => {
+        syncTime(timezone);
+      }, 1000);
+
+      return () => clearInterval(interval);
+    }
+  }, [editable]);
+
+  const handleChange = (type: "hour" | "minute" | "second", value: string) => {
```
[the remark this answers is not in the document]
Alimnfl: done

adamnoto (approved):

PR #12, opened 2025-07-05
[Feature] WORKATY-81: Story Book: Part 7

Alimnfl:


adamnoto (commented):
It's still work in progress, but can't wait to give an early feedback so here they are :)

1. For the button, can we use `#bdbdbd` for the divider when there's tip menu defined?

    To indicate which button trigger the tipmenu, like in toolbar, can we set the tipmenu toggler button to grey when the tipmenu is shown? regardless of whether the user is hovering over that tipmenu shower button or not.

    Screenshot 2025-07-05 at 23 59 28

2. For the table, when showing empty slate, can we also render the border, so that we know the empty slate is talking about the table, and that indeed the table has no content to be shown

    Screenshot 2025-07-06 at 0 04 14

components/table.stories.tsx
```diff
@@ -51,47 +166,280 @@ export const TableDefault: Story = {
...
+            rowId={`${rowValue.from}-${rowValue.content}`}
+            actions={ROW_ACTION}
+          >
+            {[rowValue.from, rowValue.content].map((rowCellValue, i) => (
+              <Table.Row.Cell
+                width={columns[i].width}
+                className={i === columns.length - 1 ? "pr-9" : ""}
+                key={`${rowValue.from}-${rowValue.content}-${rowCellValue}`}
+              >
+                <span dangerouslySetInnerHTML={{ __html: rowCellValue }} />
```
adamnoto, line 380: can we render <strong> and things manually without using `dangerouslySetInnerHTML` ? let's never use this kind of strong/hacky method :)

components/table.stories.tsx
```diff
@@ -51,47 +166,280 @@ export const TableDefault: Story = {
...
+      >
+        {rows.map((rowValue, rowIndex) => (
+          <Table.Row
+            key={rowIndex}
+            rowId={`${rowValue.from}-${rowValue.content}`}
+            actions={ROW_ACTION}
+          >
+            {[rowValue.from, rowValue.content].map((rowCellValue, i) => (
+              <Table.Row.Cell
+                width={columns[i].width}
```
adamnoto, line 376: currently we set width manually like this, and this won't be convenient and very manual to set. may I suggest we use proper `<table>` ? which then, we can set [the header](https://jsfiddle.net/mz6h2abn/) I'd rather we use table, than using hacky technique, or pass in style manually with logic. do you think there's a drawback from using a real `<table>`?

components/treelist.stories.tsx
```diff
@@ -40,12 +44,127 @@ export const Default: Story = {
         <TreeList
           content={TREE_LIST_DATA}
           emptySlate={
-            <div className="text-sm font-semibold w-full items-center flex">
-              Not found.
-            </div>
+            <EmptySlate
```
adamnoto, line 47: I don't know why need this here since we don't use `emptySlate` anyway... I think if we want to demonstrate this `emptySlate` then we need to create a new story with no data

components/rich-editor.tsx
```diff
@@ -0,0 +1,344 @@
...
+    onChange?.(markdown);
+  };
+
+  const handleOnKeyDown = (e: KeyboardEvent<HTMLDivElement>) => {
+    if ((e.ctrlKey || e.metaKey) && e.key.toLowerCase() === "u") {
+      e.preventDefault();
+      return;
+    }
+
+    if (e.key === " ") {
```
adamnoto, line 78: can you please add a comment why there is need for this, sorry it's quite hard to understand

also, I found this bug:

https://github.com/user-attachments/assets/8ca7ac4e-f1c7-4832-87ca-58eeaebfa57f

components/calendar.tsx
```diff
@@ -315,58 +315,61 @@ export default function Calendar({
...
+                onChangeValueDate({
+                  target: { name: "year", value },
+                });
+              }}
+            />
+            <button
+              onClick={() => handleClickMode("open")}
+              className="bg-white text-black rounded-xs px-[6px] text-[10px] py-1 focus:outline-none max-h-[34px] font-extrabold focus:ring-2 focus:ring-[#61A9F9] cursor-pointer hover:bg-gray-200"
+            >
+              ✔
```
adamnoto, line 369: sorry, can we use `ri-check-line` remix icon again? just the icon needs to be change, the button shape and behavior in terms of consistency with the `>` and `<` is alright

components/combobox.stories.tsx
```diff
@@ -117,3 +118,97 @@ export const WithClearable: Story = {
...
+    await expect(grapeOption).toBeVisible();
+
+    await userEvent.keyboard("{arrowdown}");
+    await userEvent.keyboard("{enter}");
+
+    await expect((input as HTMLInputElement).value).toBe("Grape");
+  },
+};
+
+export const WithStrictValue: Story = {
```
adamnoto, line 184: just `StrictValue` is enough

components/empty-slate.tsx
```diff
@@ -33,7 +33,7 @@ export default function EmptySlate({
         </div>
       )}
       <div className="flex flex-col gap-1 items-center">
-        <h2 className="font-semibold text-xl">{title}</h2>
+        <span className="font-semibold text-xl">{title}</span>
```
adamnoto, line 36: I think this one using `<h2>` is appropriate

components/table.tsx
```diff
@@ -32,22 +55,44 @@ interface TableProps {
   setIsOpen?: () => void;
   subMenuList?: (columnCaption: string) => TipMenuItemProps[];
   emptySlate?: ReactNode;
+  onLastRowReached?: () => void;
+  showPagination?: boolean;
+  onNextPageRequested?: () => void;
+  onPreviousPageRequested?: () => void;
+  pageNumberText?: number;
```
adamnoto, line 62: let's make this string, we don't assume how the page text is rendered, user can do `pg. {pageNumber}` or `hal. {pageNumber}` so to offer that flexibility, let's make this a string and render whatever is given by the user

components/table.tsx
```diff
@@ -32,22 +55,44 @@ interface TableProps {
...
+  disablePreviousPageButton?: boolean;
+  disableNextPageButton?: boolean;
+  allRowIds?: string[];
```
adamnoto, line 65: I think we can remove this

adamnoto (changes requested):
awesome work. just "some" comments and request :)

also... can we also remove unnecessary dependencies, I don't think we need `axios`, `shadcn`, `libphonenumber-js` etc, please remove as much as possible and sensible of our dependencies

also... let's rename `Button.stories.tsx` to `button.stories.tsx` for consistency

also... on stories, can we not using alert? maybe, can we just like add <p> with certain ID on the `document` or `body` of the story, and then assert the existence of the <p>. because `alert` is so distracting :(

components/calendar.tsx
Alimnfl: done, but you must see again about check line if we use on 24 px, I think like not same on chevron, wdyt?

components/treelist.stories.tsx
Alimnfl: done, i make this use WithEmptySlate, this is same like label

components/table.tsx
Alimnfl: done

components/table.tsx
Alimnfl: done

components/table.stories.tsx
Alimnfl: done

components/empty-slate.tsx
Alimnfl: done

components/table.stories.tsx
Alimnfl: done

components/rich-editor.tsx
Alimnfl: > can you please add a comment why there is need for this, sorry it's quite hard to understand
> 
> also, I found this bug:
> 
>  Screen.Recording.2025-07-15.at.6.40.23.AM.mp4

so for this one is if we enter the space like ordered list and unordered list, have some list after that, for bug i'll fix for now

adb819tektome (dismissed):

PR #15, opened 2025-07-17
[Feature] WORKATY-84: Create <DormantText> component

alimnfl:


components/paper-dialog.tsx
```diff
@@ -60,12 +60,7 @@ const PaperDialogBase = forwardRef<PaperDialogRef, PaperDialogProps>(
       setDialogState(open);
 
       controls.start({
-        x:
```
adamnoto, line 63: I can tolerate this change in this PR :) future note: this is a bug fix on paper dialog when left-side mode was buggy.

components/textbox.tsx
```diff
@@ -91,22 +94,47 @@ const Textbox = forwardRef<
...
     const inputClass = cn(
-      "rounded-xs border text-black px-2 w-full py-[7px] outline-none",
+      "rounded-xs border text-xs text-black px-2 w-full py-[7px] outline-none",
       showError
         ? "border-red-500 focus:border-red-500 focus:ring-red-500 text-red-800"
         : "border-gray-300 focus:ring-[#61A9F9] focus:border-[#61A9F9]",
       className
     );
 
+    const iconSizeDormanted = dormantedFontSize * 1.05;
```
adamnoto, line 104: `dormantPencilSize` ... or since this is not used anywhere else, let's just set it on the `size` prop on the component?

package.json
```diff
@@ -249,6 +249,7 @@
     "type": "git",
     "url": "https://github.com/systatum/coneto"
   },
+  "homepage": "https://coneto.systatum.com",
   "author": "Adam Noto Hakarsa",
```
adamnoto, line 253: just realized this, let's just write Systatum here

package.json
```diff
@@ -249,6 +249,7 @@
...
   "license": "MIT",
   "keywords": [
```
adamnoto, line 255: well, if we can add "coneto" might as well add "systatum" :)

adamnoto (changes requested):
looks good, but let's have quick talk about the animation

package.json
Alimnfl: done

components/textbox.tsx
Alimnfl: done

components/paper-dialog.tsx
Alimnfl: oh god, sorry

components/dormanted-text.tsx
```diff
@@ -0,0 +1,111 @@
+import {
+  RemixiconComponentType,
+  RiCheckLine,
+  RiPencilFill,
+} from "@remixicon/react";
+import { ReactNode, useState } from "react";
+import { cn } from "../lib/utils";
+
+export interface DormantedTextProps {
```
adamnoto, line 9: can we rename it to `DormantText` please :)

components/dormanted-text.tsx
```diff
@@ -0,0 +1,111 @@
...
+  label?: string;
```
adamnoto, line 10: hm. is this used? if not let's remove this

components/dormanted-text.tsx
```diff
@@ -0,0 +1,111 @@
...
+      const height = el.getBoundingClientRect().height;
+      if (el instanceof HTMLLabelElement) {
+        setLabelHeight(height);
+      } else {
+        setInputHeight(height);
+      }
+    }
+  };
+
+  console.log(inputHeight);
```
adamnoto, line 46: not needed

components/paper-dialog.tsx
adamnoto: yeah it can be inconsistent but ideally let's try one PR for one topic if possible, rare exception is when the topic is about adding many things all at once

adamnoto (changes requested):
Request behavior change. So, for example, if we edit a dormanted text, the editor (be it input or combo) should be having the same width as when it is dormanted. It may sound confusing, let's have a discussion later.

components/dormanted-text.tsx
Alimnfl: okay, done

components/dormanted-text.tsx
Alimnfl: done

components/dormanted-text.tsx
Alimnfl: done

adamnoto (changes requested):
Thanks Alim, but there seems to be bug for example when clicking the last example on [this story](https://deploy-preview-15--coneto.netlify.app/?path=/story/stage-dormanttext--with-moneybox).

Screenshot 2025-07-23 at 22 45 10

components/dormant-text.tsx
```diff
@@ -74,7 +74,7 @@ function DormantText({
 
     if (type === "selectbox") {
       setLabelWidth((prev) => prev + 30);
-    } else if (type.startsWith("timebox")) {
+    } else if (typeof type === "string" && type.startsWith("timebox")) {
```
adamnoto, line 77: what if instead of doing this, on line 73, we automatically convert `type` to `String(...)` ?

adamnoto (changes requested):

components/dormant-text.tsx
Alimnfl: This is automaticly string, but TypeScript must have double check on useEffect, but I've alternative code how to declare

  useEffect(() => {
    if (!inputRef.current) return;

    const type = String(inputRef.current.dataset.type);

    if (type === "selectbox") {
      setLabelWidth((prev) => prev + 30);
    } else if (type.startsWith("timebox")) {
      if (type === "timebox-with-second") {
        setLabelWidth((prev) => prev + 52);
      } else {
        setLabelWidth((prev) => prev);
      }
    }
  }, [dormantedLocal]);

adamnoto (approved):

PR #17, opened 2025-07-19
[Bug] WORKATY-87: Fix buggy auto-scroll on combobox

Alimnfl:


adamnoto (approved):
Let's add a test ensuring it doesn't auto-scrolling when we have added playwright/etc.

PR #18, opened 2025-07-19
[Bug] WORKATY-89: Fix calendar bug dates after clicking today

Alimnfl:


adamnoto (approved):
Let's add a test ensuring can travel and use future-future months, and past-past months later when we have added playwright/etc.

PR #20, opened 2025-07-23
[Chore] WORKATY-91: Migrate from test storybook to cypress on Coneto

alimnfl:


components/avatar.tsx
```diff
@@ -71,6 +71,9 @@ function Avatar({
       {...props}
       onClick={changeable ? handleClick : props.onClick}
       onChange={changeable ? handleFileChange : onChange}
+      data-testid="avatar-content"
```
adamnoto, line 74: let's try avoiding using `data-testid` as much as possible? perhaps, we can make this element to receive `id` props, and then set the `id` and then we can find the element through that ID.

indeed, setting the ID manually is not possible at all time, but we should be able to find a way, I will discuss below for those cases.

components/boxbar.tsx
```diff
@@ -36,17 +36,15 @@ function Boxbar({
       <span className={cn("p-2 w-full flex-wrap flex", childClassName)}>
         {children}
       </span>
-      <button
+      <motion.button
+        data-testid="boxbar-toggle"
```
adamnoto, line 40: do you think it's possible to give it a class name and use that class name instead?

maybe something like:

```scss
.boxbar-toggle {
  @apply mt-2 mr-2 top-2 w-fit right-2 cursor-pointer hover:bg-gray-100 p-1 rounded;
}
```

Then uses it:

```ts
<motion.button
  className="boxbar-toggle"
  animate={{ rotate: isOpen ? 90 : 0 }}
  transition={{ duration: 0.2 }}
  onClick={() => setIsOpen(!isOpen)}
>
  {/* ... */}
</motion.button>
```

So we can target by the ID and then the class name. Basically avoiding using `data-testid` again

components/moneybox.tsx
```diff
@@ -89,6 +89,8 @@ const Moneybox = forwardRef<HTMLInputElement, MoneyboxProps>(
       className
     );
 
+    console.log(currency);
```
adamnoto, line 92: let's remove console.log :)

components/toolbar.tsx
```diff
@@ -178,12 +178,16 @@ function ToolbarMenu({
   );
 
   return (
-    <div ref={containerRef} className="relative flex flex-col mr-1">
+    <div
+      data-testid={`toolbar-menu-${caption}`}
```
adamnoto, line 182: I think we should be able to find by the caption/text without setting it as `data-testid`

components/toolbar.tsx
```diff
@@ -178,12 +178,16 @@ function ToolbarMenu({
...
+      ref={containerRef}
+      className="relative flex flex-col mr-1"
+    >
       <div ref={refs.setReference} className={toolbarMenuClass}>
         {(Icon || caption) && (
           <>
-            <div
-              aria-label="toolbar-menu-toggle"
+            <button
+              aria-label={`toolbar-menu-button-${caption}`}
```
adamnoto, line 190: is it really necessary?

components/window.tsx
```diff
@@ -149,6 +149,7 @@ function Window({
 
           {index < childrenArray.length - 1 && (
             <div
+              data-testid={`window-divider-${index}`}
```
adamnoto, line 152: I think we can give id to the main element, and then target the n-th children either using CSS or xpath

components/window.tsx
```diff
@@ -187,9 +188,12 @@ function WindowCell({ children, className, actions }: WindowCellProps) {
                 )}
               >
                 {data.icon && (
-                  <span className="absolute hover:bg-gray-400 transition-all duration-300 right-2 top-1/2 -translate-y-1/2">
+                  <button
+                    data-testid="window-button"
```
adamnoto, line 192: similarly, let's use xpath perhaps?

test/e2e/dormant-text.cy.ts
```diff
@@ -0,0 +1,129 @@
...
+      cy.findAllByText("Orange").should("have.length.at.least", 1);
+    });
+  });
+
+  describe("Datebox", () => {
+    beforeEach(() => {
+      cy.visit(getIdContent("stage-dormanttext--with-datebox"));
+    });
+
+    it("Should open date picker and click a date", () => {
```
adamnoto, line 61: let's also test and ensure we have the expected text/date in the dormant text; let's do the same for other examples (ie phonebox, etc) of which the shown/displayed value we haven't assert.

test/e2e/rich-editor.cy.ts
```diff
@@ -0,0 +1,46 @@
+import { getIdContent } from "test/support/commands";
+
+context("RichEditor Component", () => {
```
adamnoto, line 3: can I suggest we chagne "Print Console" to just "Print" and print that to a `<p>` below the editor? and then, we can test the represented content through this `<p>`...

adamnoto (changes requested):
Very nice, just small request. Also, let's add in the README.md about how we can run the test.

components/toolbar.tsx
Alimnfl: example

 <button
          aria-label={`toolbar-menu-toggle`}
          onMouseEnter={() => setHovered("dropdown")}
          onMouseLeave={() => setHovered("original")}
          className={cn(
            "flex p-2 relative h-full items-center max-w-[36px]",
            hovered === "dropdown" && toolbarMenuHoverClass,
            isOpen && toolbarMenuBackgroundActiveClass,
            isOpen && toolbarMenuBorderActiveClass,
            triggerClassName,
            isOpen && toggleActiveClassName
          )}
          onClick={handleClickOpen}
        >
        
        wdyt if i'll use with aria-label? and not use with data-testid?

components/window.tsx
Alimnfl: I'll use with aria-label, this is easy to use for test component.

components/window.tsx
Alimnfl: I'm not using indexing for now, we can detect elements with Cypress anyway. Sorry for the oversight, I'll use aria-label instead.

components/toolbar.tsx
Alimnfl: I'm not use with ${captions} for now, this is not necessary

adamnoto (approved):

PR #28, opened 2025-07-31
[Improvement] WORKATY-103: Tooltip text using ReactNode for customize and have arrow style

alimnfl:


components/tooltip.stories.tsx
```diff
@@ -109,9 +126,86 @@ export const BlueUnderlineTooltip: Story = {
...
     return <Tooltip {...args} />;
   },
 };
+
+export const Custom: Story = {
+  render: () => {
+    const [value, setValue] = useState({
+      division_name: "",
+    });
+    const onChangeDivisionEmployeeForm = (
```
adamnoto, line 142: I think now we can use the simpler type definition here

adamnoto (changes requested):

PR #31, opened 2025-08-01
[Bug] WORKATY-105: Remove unnecessary style on DrawerTab

alimnfl:
`DrawerTab` now uses proper fixed positioning on the left or right, so the additional `wrapper/container` and default bg-white are no longer needed.

Screenshot 2025-08-01 at 17 26 01


adamnoto (approved):
yay for all that's simpler and better

PR #41, opened 2025-08-14
[Improvement] WORKATY-121: Add onclick on table row

Alimnfl:
For this PR, I add props onClick for `TableRow` and `TableRowCell`, you can check for new behaviour on story appendable to click on row, and the result you can see also on mode dev.

### DIFF
--- a/components/table.stories.tsx
+++ b/components/table.stories.tsx
@@ -376,11 +376,12 @@ export const Appendable: Story = {
       >
         {rows.map((rowValue, rowIndex) => (
           <Table.Row
-            onClick={() =>
+            onClick={({ setIsChecked }) => {
               console.log(
                 `Selected to this ${`${rowValue.from}-${rowValue.content}-${rowValue.subject}`}`
-              )
-            }
+              );
+              setIsChecked(true);
+            }}
             key={rowIndex}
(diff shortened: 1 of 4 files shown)


components/table.stories.tsx
```diff
@@ -376,12 +376,21 @@ export const Appendable: Story = {
       >
         {rows.map((rowValue, rowIndex) => (
           <Table.Row
+            onClick={() =>
```
adamnoto, line 379: let's add capability to check the row (when it has a checkbox):

```ts
onClick={({setIsChecked}) => {
  /* make the row checked; but if the row doesn't have a 
     checkbox, calling setIsChecked will not do anything, 
     please ensure this behavior in a test */
  setIsChecked(true);
})
```

this way, we don't assume that clicking the row will trigger checking -- that'd be user-defined behavior.

please make the example "Table / Sortable With Pagination" modified such that clicking the row will also set the checkbox of the row to `checked`.

adamnoto (changes requested):
adding feature to check the checkbox

adamnoto:
@Alimnfl Hmm I am curious, can Cypress be used to test that an error is not raised? I am interested that we add a test case whereby we render table without checkbox, but add row onclick with togglecheckbox, and calling togglecheckbox shouldn't break the table.

adamnoto (approved):

PR #44, opened 2025-08-15
[Improvement] WORKATY-118: Add missing fields and other props on the StatefulForm

Alimnfl:
I’ve added a few new things to the togglebox, new props, an updated story and test, an exported interface, and label styling.

### DIFF
--- a/components/rating.tsx
+++ b/components/rating.tsx
@@ -1,12 +1,19 @@
-import { MouseEvent, useState } from "react";
-import styled, { css } from "styled-components";
+import { ChangeEvent, MouseEvent, ReactElement, useState } from "react";
+import styled, { css, CSSProp } from "styled-components";
 
 export interface RatingProps {
-  rating?: number;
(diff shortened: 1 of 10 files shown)


components/rating.tsx
```diff
@@ -1,12 +1,19 @@
-import { MouseEvent, useState } from "react";
-import styled, { css } from "styled-components";
+import { ChangeEvent, MouseEvent, ReactElement, useState } from "react";
+import styled, { css, CSSProp } from "styled-components";
 
 export interface RatingProps {
-  rating?: number;
-  onChange?: (rating: number) => void;
+  rating?: string;
```
adamnoto, line 5: why do we change it to `string` ? ... if we must use `string` because I am seeing down below we are doing `number()` which then... why don't we just use number from the get go?

Alimnfl: Because on Change Event HTML Input must using with string

adamnoto (approved):

PR #45, opened 2025-08-16
[Bug] WORKATY-126: Fix slight corner border styling bug on CapsuleTab

Alimnfl:
For now, it's using with 2px on radius from on before 0.375rem (6px), wdyt?


Alimnfl:
For now, it's using with 2px on radius from on before 0.375rem (6px), wdyt?

PR #47, opened 2025-08-16
[Improvement] WORKATY-131: Add export interface for DrawerTab, CapsuleTab, and other component.

Alimnfl:
Let's add some props to helping on the future, if we want to make constant data for content like on drawer, capsule tab.


adamnoto (changes requested):
nice, please help with resolving the conflict first

adamnoto (approved):

PR #48, opened 2025-08-19
[Bug] SYST-137: Fix bug combo may not be highlighted when going backward

Alimnfl:
image

wdyt for that, for now we have 2 choice to fix this, for the first if we want have combobox still have on 33 after refresh on combobox (assumption now we have value on 50) just have one way to make consistent, just using like localstorage (that's not affect with any state),
 
but we have 1 solution option also, it's like if we have value same like before ( I mean at 50), we can make the value of combo it's using with clamp (that's the last number of combo -> we can add using useEffect to make this)...

### DIFF
--- a/components/pagination.tsx
+++ b/components/pagination.tsx
@@ -1,7 +1,7 @@
 import { RiArrowLeftSLine, RiArrowRightSLine } from "@remixicon/react";
 import { Combobox } from "./combobox";
 import { OptionsProps } from "./selectbox";
-import { ReactNode, useEffect, useState } from "react";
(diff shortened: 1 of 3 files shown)


components/pagination.tsx
```diff
@@ -47,6 +48,12 @@ function Pagination({
     }
   };
 
+  useEffect(() => {
+    const comboboxPagesNumber = totalPages - 3;
```
adamnoto, line 52: why do we need to define `comboboxPagesNumber` again? it's been defined on line 29 right?

Alimnfl: Okay, I'll remove it

PR #61, opened 2025-08-29
[Improvement] SYST-158: Give depth when clicking a button

Alimnfl:


Alimnfl:
> Quick change: When the tip menu opens from a dropdown button, please keep that button’s background the same as its pressed state while hovering over the menu.

okay done, you can check it new update.

adamnoto (changes requested):
On button which can reveal tip menu, when the button is clicked, and tip menu is revelaed, let's keep the inlet/shadowing effect like when it's pressed. Please apply this to any button-like element that can have tipmenu (ie. toolbar)

test/component/toolbar.cy.tsx
```diff
@@ -0,0 +1,153 @@
...
+              subMenuList={TIP_MENU_ITEMS}
+              dropdownStyle={{ minWidth: "240px" }}
+            />
+          </Toolbar>
+        );
+
+        cy.findByLabelText("toolbar-menu-toggle")
+          .realHover()
+          .wait(300)
+          .should("have.css", "background-color", "rgb(245, 245, 245)"); // sesuai VARIANT_CLASS_MAP.hover.default (#f5f5f5)
```
adamnoto, line 149: let's change comment language

adamnoto (changes requested):
just a bit more, looks good :)

PR #63, opened 2025-09-02
[Improvement] SYST-165: Add ability to add tip menu on table action buttons

Alimnfl:


Alimnfl:
> Let's render the `|` separator with border `1px solid rgb(236 236 236)`. Also let's introduce `size` `xs` to `button` to make them way smaller again. Also, let's introduce `size` `sm` to tipmenu (default is `md` which is the current one), and so reduce the padding and the sizing so the menu is smaller too, but still readable.

For now, it's still on separator just on table right, but on introducing it on button our component, was done for now, on the smaller size had finished.

components/button.tsx
```diff
@@ -34,6 +35,9 @@ function Button({
   containerStyle,
   buttonStyle,
   toggleStyle,
+  onClick,
+  dividerStyle,
+  tipMenuVariant,
```
adamnoto, line 40: let's call it `tipMenuSize`

adamnoto (changes requested):
a bit more :D

components/button.tsx
Alimnfl: done

PR #72, opened 2025-09-11
[Improvement] SYST-169: Ability to select a range of date for Calendar via selectability

alimnfl:


adamnoto (changes requested):
🔥 🚒

adamnoto (changes requested):
A bit more :0

Screenshot 2025-09-14 at 0 53 30

Please when select in range, if the start date > today's date, let's turn the today's date text color in gray like all the other dates in the past.

Also could we please add unit tests for `Calendar` to ensure the `value` format for all different `selectability`, ie when ranged it should be in the `date1-date2` form, when multiple selectability, it should be `date1,date2,...date_n`, else by default just `date`.

Alimnfl:
> A bit more :0
> 
> Screenshot 2025-09-14 at 0 53 30
> Please when select in range, if the start date > today's date, let's turn the today's date text color in gray like all the other dates in the past.
> 
> Also could we please add unit tests for `Calendar` to ensure the `value` format for all different `selectability`, ie when ranged it should be in the `date1-date2` form, when multiple selectability, it should be `date1,date2,...date_n`, else by default just `date`.

there was done, for ensure also with formatted on selectability mode `single`, `multiple`, and `ranged`.

adamnoto (changes requested):
There's visual regression on the dot. The dot before this change for today always in the center, but now it's no longer be the case 🙏 

In this branch:

Screen Shot 2025-09-19 at 9 38 20 PM

Screen Shot 2025-09-19 at 9 39 53 PM


In `main`:

Screen Shot 2025-09-19 at 9 40 01 PM

Do you think it's possible to add a unit test to ensure the circle dot is in the middle?

adamnoto (changes requested):
Nice, just a bit more test expectation request 🙂

adamnoto (changes requested):
One more: please make today's date gray if it falls on weekend, and the calendar is set to be no weekend. Please add the test too :) ensuring color is gray in such condition.

adamnoto (changes requested):
Currently when it's no weekend and today's date is on weekend, this would happen. Could we please don't select the date in case it's weekend and non-weekend? This happens [in this story](https://deploy-preview-72--coneto.netlify.app/?path=/story/input-elements-calendar--no-weekends) but I think the bug might have been on the `Calendar` itself. Please fix and write the test for the fix :) ensuring it should not happen again.

Screenshot 2025-09-21 at 12 22 28

adamnoto (approved):
dapuran

PR #74, opened 2025-09-16
[Feature] SYST-184: Pinbox

Alimnfl:


adamnoto (changes requested):
1. Let's standardize the font-size, by default to 24px. Keep do the underscore indicator calculation, and the height-width calculation correctly and adequately. Maybe it's no longer 80% for the underscore indicator. But yeah, let's have a bigger and consistent `font-size` for this `Pinbox`.
2. Let's always have textbox for each part of the text (even static).
3. Let's have margin/padding (whichever is fine) of around ~4px.

Other than that, cool 👍 . Please add some tests for all the behavior.

adamnoto (changes requested):
Please render the static item/text also inside a textbox. I think this is all that's missing/haven't been implemented 🙏

Alimnfl:
> Please render the static item/text also inside a textbox. I think this is all that's missing/haven't been implemented 🙏

done

PR #80, opened 2025-09-20
[Improvement] SYST-167: Add mode and toolbarPosition to RichEditor

Alimnfl:


adamnoto (changes requested):
Nice! And thanks for about writing the test.

⚠️  That being said, as per task `mode` is `text-editor`, `page-editor`, or `view-only`. Please take a look at the task again and cross-check, because it's still a bit different from the task. 🙂 

The task should be complete and self-describing. But, please let me know if you have any question. All ears.

adb819t (changes requested):
All good, I will withhold approving tho since I believe we'd investigate making Ctrl+C (or Cmd+C) works when on the `view-only` mode.

adb819t (approved):

adb819t:
40521256-9252103e-6005-11e8-9ba6-eeb9c15bda68

adb819t (approved):

adamnoto (approved):

Alimnfl:
> All good, I will withhold approving tho since I believe we'd investigate making Ctrl+C (or Cmd+C) works when on the `view-only` mode.

yeps, absolutely using this one right now 🙌

PR #86, opened 2025-09-26
[Improvement] SYST-214: Generalize action items on Badge

Alimnfl:


adb819t (changes requested):
Q: is it possible not to use state for hover, just use browser's built-in `:hover` detection mechanism...? If possible, can we remove it, I think then we can remove `styleWithProp` too :)

Alimnfl:
> > Q: is it possible not to use state for hover, just use browser's built-in `:hover` detection mechanism...? If possible, can we remove it, I think then we can remove `styleWithProp` too :)
> 
> why we should use it? because we can't catch for hover inside of component actually, but I'll try this one (so tricky :) )

I have solution for now, we can catch with aria-label for that, Alhamdulillah :)

adb819t:
nice, yeah, let's use that instead of state :)

adb819t (approved):
menarik

adamnoto (approved):

PR #91, opened 2025-10-03
[Improvement] SYST-223: Generalize chip functionality so it can contain any form

Alimnfl:


adamnoto (changes requested):
1. please also fix on the stateful form, where it's not using the parent/form maximum size
2. is the chip zod object still makes sense on the stateful form? I thought we shouldn't have this form anymore:

   ```js
      chips: z.object({
        search: z.string().optional(),
        name_tag: z.string().optional(),
        background_color: z.string().optional(),
        text_color: z.string().optional(),
        circle_color: z.string().optional()
      }),
   ```

    please refer to the task for the expected form, wdyt?

components/chips.stories.tsx
```diff
@@ -102,6 +104,121 @@ export const Default: Story = {
       console.log("clicked new tag");
     };
 
+    const MissingContent = ({
```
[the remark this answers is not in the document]
Alimnfl: done

adb819t (changes requested):

components/chips.tsx
```diff
@@ -598,7 +496,7 @@ const ChipsDrawerWrapper = styled.ul<{
   ${({ $style }) => $style}
 `;
 
-const NewTagOption = styled.div<{ $hovered: boolean }>`
+const EmptyOption = styled.div<{ $hovered?: boolean }>`
```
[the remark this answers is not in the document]
Alimnfl: okay done

components/chips.stories.tsx
```diff
@@ -102,9 +106,124 @@ export const Default: Story = {
       console.log("clicked new tag");
     };
 
+    const MissingOptionForm = ({
+      inputRef,
```
adamnoto, line 110: let's call this `firstFocusRef` ?

components/chips.stories.tsx
```diff
@@ -102,9 +106,124 @@ export const Default: Story = {
...
+      switchToIdle,
```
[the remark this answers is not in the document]
Alimnfl: done

components/chips.stories.tsx
Alimnfl: so for this one actually for the second input when we want to focusing, after switching to `creatable` mode.

components/chips.tsx
```diff
@@ -56,13 +52,26 @@ interface BaseChipsProps {
...
+  missingOptionLabel?: string;
   creatable?: boolean;
   onOptionClicked?: (badge: BadgeProps) => void;
-  selectedOptions?: number[];
-  onNewTagCreated?: () => void;
+  selectedOptions?: BadgeProps[];
+  missingOptionForm?:
+    | ReactNode
+    | ((props?: MissingOptionFormProps) => ReactNode);
+  missingEmptyContent?: ReactNode;
```
adb819t, line 62: let's just call this `emptySlate` for consistency with other components

components/chips.tsx
```diff
@@ -56,13 +52,26 @@ interface BaseChipsProps {
...
+  renderer?: (props: ChipRendererProps) => ReactNode;
+}
+
+export interface ChipRendererProps {
+  id?: string;
+  caption?: string;
+  metadata?: Record<string, unknown>;
```
adb819t, line 69: why `unknown` not `any`?

Alimnfl: CMIIW, so for unknown on typescript it will be safe instead of using any, like must to force validate type first on unknown...

components/chips.tsx
Alimnfl: okay, for missingEmptyContent right?

adb819t (approved):
b1b34bfc9edda187af9e106f73d5873f

PR #95, opened 2025-10-04
[Improvement] SYST-231: Apply consistent styling to Pinbox when disabled

Alimnfl:


test/e2e/pinbox.cy.ts
```diff
@@ -1,7 +1,7 @@
 import { getIdContent } from "test/support/commands";
 
 describe("Pinbox", () => {
-  context("Default", () => {
+  context("default", () => {
```
adb819t, line 4: should be `"when default"` context is either: `with` or `when`.

adb819t (changes requested):
all good just small change request

test/e2e/pinbox.cy.ts
Alimnfl: done

adamnoto (approved):
finnalzo

PR #97, opened 2025-10-04
[Improvement] SYST-230: Remove bolder font-weight on FileDropBox

Alimnfl:


components/file-drop-box.tsx
```diff
@@ -340,7 +340,7 @@ const IconsRow = styled.div`
 `;
 
 const PlaceholderText = styled.span<{ $isDragging: boolean }>`
-  font-weight: 600;
+  font-weight: 400;
```
adamnoto, line 343: why not just remove it?

adamnoto (commented):
...

components/file-drop-box.tsx
adamnoto: really curious... because I thought we never specify `font-weight` unless we need to?

Alimnfl: okay, I've remove it

adamnoto (approved):

PR #111, opened 2025-10-28
[Improvement] SYST-250: Allow rendering custom component on StatefulForm

alimnfl:
Tick what you have done:

[x] I have tested it myself


adamnoto (changes requested):
Let's change some stories. Maybe let's change [this](https://deploy-preview-111--coneto.netlify.app/?path=/story/input-elements-statefulform--default) and render terms and conditions with lorem ipsum text, without title, and then rename the checkbox to "I agree" instead of "Has access to login"

adamnoto (approved):
finncxqa

PR #116, opened 2025-10-30
[Improvement] SYST-252: Add hidden prop to stateful form fields

alimnfl:
Tick what you have done:

[x] I have tested it myself


adamnoto (changes requested):
Let's change a bit.. and how about adding "Middle Name" in the form, and make it hidden?

adb819t (approved):
finnasdjd

adamnoto (approved):

PR #146, opened 2025-11-15
[Bug] SYST-296: Show the datebox value when initialize from outside of component

alimnfl:
Description:
For now, we can't initialize the value from outside of component. The selectbox doesn't respond because it currently filters strictly based on the avalivable `options`

So, we must add this text with `isValidDateString`:
   ```tsx
   const initialState = options.find(
      (opt) => opt.value === selectedOptions?.[0]
    ) ?? {
      text: isValidDateString(selectedOptions?.[0]) ? selectedOptions?.[0] : "",
      value: "0",
    };
```

This PR includes several improvement related to how calendar handles and displays values:
1. Handle to show the first date when initialize on the `calendar`
2. Update the filtering logic for `single | multiple | ranged` on the level calendar, including when disableWeekend.
3. Improves value setup on the selectbox show the initial value.
4. Can add some placeholder to the selectbox.

Result:
image

Source:
[[Bug] SYST-296: Show the datebox value when initialize from outside of component](https://linear.app/systatum/issue/SYST-296/show-the-datebox-value-when-initialize-from-outside-of-component)

Tick what you have done:

[x] I have tested it myself

### DIFF
--- a/constants/valid-date.ts
+++ b/constants/valid-date.ts
@@ -1,0 +1,31 @@
+export function isValidDateString(value: string): boolean {
+  if (!value) return false;
+
+  const parts = value.includes(",")
+    ? value.split(",").map((s) => s.trim())
+    : value.includes("-")
+      ? value.split("-").map((s) => s.trim())
+      : [value];
+
+  return parts.every((p) => !isNaN(new Date(p).getTime()));
+}
+
+export function isValidSingleDate(v: string): boolean {
+  return !isNaN(new Date(v).getTime());
+}
+
+export function getValidMultipleDate(raw: string): string[] {
+  if (!raw) return [];
+  return raw
+    ?.split(",")
+    ?.map((s) => s.trim())
+    ?.filter(isValidSingleDate);
+}
+
+export function removeWeekend(dates: string[]): string[] {
+  return dates.filter((d) => {
+    const dd = new Date(d);
+    const day = dd.getDay();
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: constants/valid-date.ts [hunk 1]
```diff
@@ -0,0 +1,31 @@
+export function isValidDateString(value: string): boolean {
```
adamnoto: they are all not constant but functions, please put them inside of `lib` and rename this file to `date.ts` to make it more universal

test/component/datebox.cy.tsx
```diff
@@ -0,0 +1,161 @@
...
+            "background-color",
+            "rgb(97, 169, 249)"
+          );
+        });
+      });
+    });
+  });
+
+  context("ranged", () => {
+    context("when initialize first", () => {
```
adamnoto: `when initialized with selected date`, please remove others up above similarly

adamnoto (changes requested):

test/component/datebox.cy.tsx
```diff
@@ -0,0 +1,161 @@
+import { Datebox } from "./../../components/datebox";
+
+describe("Datebox", () => {
+  context("when initialize first", () => {
+    it("render the value on input", () => {
```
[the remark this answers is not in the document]
Alimnfl: done

adamnoto (approved):
finnzman

PR #264, opened 2026-03-04
[Bug] SYST-439: Fix bug when combobox `options` is `[]` and the value is `null`

alimnfl:
Description:
Previously, the `combobox` component when using  `selectedOptions` prop with null value, could cause a bug, as shown in the image below, because the component did not properly handle null values. This pull request fixes `combobox` to have properly filtering and always converting with string array ( `string[]` ).

```tsx
 const finalSelectedOptions = useMemo(() => {
      if (Array.isArray(selectedOptions)) {
        return selectedOptions.map(String);
      }
      if (selectedOptions != null) {
        return [String(selectedOptions)];
      }
      return [];
    }, [selectedOptions]);
```

Screen Shot 2026-02-10 at 11 34 43 AM

Source:
[[Bug] SYST-439: Fix bug when combobox options is [] and the value is null](https://linear.app/systatum/issue/SYST-439/fix-bug-when-combobox-options-is-and-the-value-is-null)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/combobox.tsx
+++ b/components/combobox.tsx
@@ -205,7 +205,7 @@ function ComboboxDrawer({
     if (Array.isArray(selectedOptions)) {
       return selectedOptions.map(String);
     }
-    if (selectedOptions !== undefined) {
+    if (selectedOptions != null) {
       return [String(selectedOptions)];
     }
     return [];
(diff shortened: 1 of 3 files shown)


adamnoto (approved):
Image

PR #326, opened 2026-03-30
[Bug] SYST-535: Controlled selection with `selectedItems` 

alimnfl:
Description:
This pull request removes unnecessary props from the `Table` component, especially at the `Table.Row` level (e.g., `isSelected`, `index`, `isLast`). These props caused ambiguity in IntelliSense and were not intendedfor public use. 

It also ensures that the `selectedItems` prop works correctly when from the parent level. This issue was identified after interacting with the `Table.Row` component, particularly due to the use of the internal `isSelected` prop.

Source:
[[Bug] SYST-535: Controlled selection with `selectedItems` ](https://linear.app/systatum/issue/SYST-535/respect-isselected-on-tablerow)

Tick what you have done:
[x] I have tested it myself


adamnoto (approved):
Image

PR #333, opened 2026-04-01
[Feature] SYST-518: Support light and dark theme.

alimnfl:
Description:
This pull request introduces a `theme` system supporting `light` and `dark` modes, with the added flexibility to fully customize themes through a configuration interface, including providers, colors, and related functionality.

Source:
[[Feature] SYST-518: Support light and dark theme.](https://linear.app/systatum/issue/SYST-518/support-theming-for-the-whole-app-with-light-being-default)

Tick what you have done:
[x] I have tested it myself


adamnoto (changes requested):
Just quick request

1. For badge, when on red variant, please use: `#3d1c1c` as text color and circle color by default.
2. Updated the spec with more coloring info for buttons, step/time line, also table. will experiment with coloring as well, but feel free to color other components in the way you feel is suitable. when you done with coloring buttons, step/time line, and table, please let me know :) wanna look and define further esp for the table.

adamedtix (changes requested):
Looks gooder and gooder :D (better is the correct word, but just wanna play around). Please apply these changes:

- [x] Default white text color for the body and other default text color for dark mode is too much, too strong, let's make it `#caced4` 
- [x] Text color in avatar should remain black, because the background color is always a bit darker.
- [x] When a checkbox is chosen/checked, both in dark and light, don't change its border color
- [x] Card's header border bottom should follow the rest's border color (should read from one same configuration)

  Image
- [x] Update choice group for radio button
  - container border color, when selected border-color: `#1465d3bf`, when unselected and not hovered: `border: 0.5px solid rgb(75, 85, 99)`, when unselected and hovered: `#1465d3bf`
  - background, when selected: `#1465d333`, when unselected: `#1f2023`

  this is quite unaccurate demo, the 100% coloring is using the above (this demo is wrong especially on border hover color when unselected, which shouldn't be gray)

  [Screencast from 2026-04-06 16-18-50.webm](https://github.com/user-attachments/assets/834c4758-86a4-4078-9f8e-1345f741d2d0)
- [x] update radio circle background to be `#1465d3bf`, but if possible, the inner circle that indicate selection should be white, instead of black, because in checkbox, the checkbox is white, not black; so let's make them uniform. the border is `1px solid rgb(55, 65, 81)`. example below is inaccurate because the inner circle is black, should be white.

  Screenshot from 2026-04-06 16-26-01

- [x] Update checkbox background and border color to: `#1465d3bf` when selected, when unselected, background is `rgb(32, 33, 35)` and border is `1px solid rgb(55, 65, 81)`. and then, when used in choicegroup, be sure to inherit background color for unselected checkboxes.

  Screenshot from 2026-04-06 16-29-40

- [x] Table's header actions container should have: `background: linear-gradient(rgb(33, 33, 33), rgb(37, 38, 38))` and `border-bottom: 0.5px solid rgb(61, 61, 61)` , and table's header should be having `background: linear-gradient(rgb(42, 42, 42), rgb(41, 44, 46))`.
- [x] Only in dark background, let's have default background color to the row group, set to: `rgb(26, 26, 26)` by default. Because, it's too blended in and because of low-light, makes it difficult to even separate rows from the row group. With all these changes, the table should look like this:

  Image

- [x] Searchbar (when dark, on the table and other places, aka the default searchbar when on dark) should have this default border: `1px solid rgb(61, 61, 61)`. when active/focus background will be: `rgb(26, 26, 26)` but keep the blueish active border color 

  Image

  Whereby the search list result should have `border: 1px solid rgb(31, 41, 55)` and background of `#1a1a1a` (please inherit from searchtext background if possible)

  Image

- [x] Make table actions have same height as the capsule, currently it's not

  Image
- [x] Treelist line when not-selected: `2px solid rgb(62, 65, 67)`. Treelist chevron background should be same with body (ie: `rgb(31, 32, 35)`, please inherit from the body)

  Screenshot from 2026-04-03 14-40-51
- [x] Unactivated toggle background should be: `rgb(80, 80, 80)` and activated: `rgb(60, 49, 110)`

  Screenshot from 2026-04-03 18-04-02

  Screenshot from 2026-04-03 18-05-01
- [x] Thumbfield when activated up: `rgb(134, 111, 238)` down: `rgb(236, 65, 108)`
- [x] Toolbar coloring (both light and dark) inherit from button's: default, primary, danger.
- [x] Tip menu, color: `#caced4` (inherit from body), bacgkround: `rgb(35, 35, 35)` also for default menu item, background: `rgb(35, 35, 35)`. `isDangerous` variant of the menu item inherits the coloring form danger button, including when hovered. for the default menu item (non dangerous), when hovered, set background color to: `#292929`

  Screenshot from 2026-04-03 18-19-21
- [x] Overlay, background color: `rgba(20, 20, 20, 0.5)`, and filter `blur(2px)`
- [x] NavTab, background: `rgb(35, 35, 35)` (inherit from `body`), background when hovered or activated: `rgba(52, 52, 52, 0.5)`. The whole container `border-bottom: 1px solid rgb(26, 26, 26)`.

  Screenshot from 2026-04-03 18-34-26
- [x] Frame should have border of: `1px solid rgba(71, 71, 71, 0.56)`, and the title background follows body's default background color.

  On the ["With frame" example of StatefulForm](https://deploy-preview-333--coneto.netlify.app/?path=/story/input-elements-statefulform--with-frame&globals=theme:dark), please render the business expense field with the background color of: `rgb(48, 48, 48)`, I think something like this can happen:

  ```tsx
  const { theme } = useTheme()
  <StatefulForm ...>
    <Frame>
      <TheContainer style={{ background: theme === "dark" ? "rgb(48, 48, 48)" : "the-original-color-idk" }}
    </Frame>
  </StatefulForm>
  ```

  Screenshot from 2026-04-05 19-09-22
- [x] When on `light` mode, selected combobox item should have white text color (currently it's black, which is a regression)

  Screenshot from 2026-04-06 15-30-41


The coloring guideline above is still experimentation, they might change, but for now, let's adopt and implement that. Feel free to edit this markdown and check the checkbox that has been implemented.

adamedtix:
On top of the [UI guidelines](https://github.com/systatum/coneto/pull/333#pullrequestreview-4054338175) above, here additional one; these should overwrites the previous guideline if there's any conflict.

- [x] For red badge, please render color and circle color to be white (it's weird using that dark red)
- [x] Remove border top on card's header, and border bottom on card's header for dark theme (I think don't remove it really, just set the border-bottom color to transparent, for example). I think it's cleaner this way, less clutter.

  <div style="display: flex; gap: 10px;">
    Screenshot from 2026-04-07 16-55-26
  
  Screenshot from 2026-04-07 16-55-35
  </div>
- [x] `DocumentViewer` background should be `#434345`, instead of `#0f172a`
- [x] When hovering for "No content" part on a list, use background color of: `#212c37` instead of `aliceblue`
- [x] Loading overlay for list (and for other similar content container, like table) should use background: `rgb(17 17 17 / 80%)`, and backdrop-filter: `blur(1px)`
- [x] Loading spinner to use color of `rgb(142 214 255)`
- [x] List with badge should have the badge background set to `rgb(13 21 25)`

  image
- [x] When toggle is active/on, use background color of `rgb(72, 57, 141)` (like when button is hovered)
- [x] Make loading skeleton color from dark blue to dark grey (but a bit light, but not so light)

adamedtix (changes requested):
Please make these changes, unless specified as `light`, please assume they are all for `dark`

- [x] When button pressed on dark, change the background. Reasoning, it's weird that when pressed, the button becomes lighter in color, although it is okay when hovered it's lighter, but when pressed, it looks better that they're darker again.
  - Primary: `rgb(60 49 110 / 80%)`
  - Danger: `rgb(167 7 43 / 80%)`
  - Default: `rgb(39 39 39 / 85%)`
  - Other (like success, etc): please find color that you feel is good, generally is 80% darker of the original color when it's not hovered aka in normal state.
- [x] When the button has tip menu, there's a `|` now, this `|` color needs to blend better.
  - In light mode: default: `#e5e7eb`, primary: `rgb(80, 145, 223)`, danger: `#a12f4b4a`, success: `rgba(19, 156, 17, 0.69)`
  - In dark mode: default `#363636`, primary: `#48398dbf`, danger: `#9315348c`, success: `rgba(19, 156, 17, 0.69)`
  - Please write a manual unit test to ensure those coloring being picked up correctly 🙏 
- [x] Update [this card default story](https://deploy-preview-333--coneto.netlify.app/?path=/story/content-card--default) to add the "Success" button
- [x] Keynote key should have `rgb(243, 244, 246)` color

  Image
- [x] On dark, when checkbox is not checked, border: `1px solid rgb(55, 59, 65)`, when checked: `1px solid rgb(21, 82, 164)`.
- [x] When hovering on list (and table row as well, and places like that), and then the `...` burger is shown, when the burger is hovered, please find an appropriate color, currently it's too light:

  Image
- [x] Don't change the hover color on list when on light background (ex in [Accordion story](https://deploy-preview-333--coneto.netlify.app/?path=/story/content-list--accordion))

  Image
- [x] When on Tooltip, ie with form that show a popup, when on dark, the background: `#292c2e`, border-color: `#3e4143`, color follows the body's background color. For the triangle, border-color: `#3e4143`, background: `#3e4143`

  Image
- [x] For TreeList on dark, the unselected line has this border-left: `2px solid rgba(63, 62, 62, 0.35)`

  Image
- [x] For capsule or capsule tab. When the tab is selected, background: `#303030`. When hovering on unselected tab, the hover capsule box should be having color of: `#303030`. The capsule tab header: `#262627`.

  Image
- [x] Please fix opening dialog and modal dialog with `.show()` call is no longer working
- [x] Dialog background is: `rgb(26, 26, 26)`, box-shadow: `rgba(0, 0, 0, 0.65) 0px 8px 24px`. Please make the circle around the icon, a bit lighter, not too dark, but also not too lightened.

  Image

- [x] For dorman text, when hovered, `border-color` must be same with `background-color`: `rgb(26, 26, 26)`

  image

- [x] Frame should have background same with body's background (also the frame's title): `rgb(31, 32, 35)`

  image
- [x] Separator text color should be: `rgb(171, 171, 171)`, line box-shadow: `rgb(43, 43, 43) 0px 2px 2px inset, rgb(130, 130, 130) 0px -1px 1px inset`. 

  image
- [x] Rename "Button > All Variants" story to be named "Default"
- [x] Switch default background coloring for `Default` to be like `Secondary`, and `Secondary` to be `Default`. Do this ONLY after you follow ALL the other guidelines above ^
- [x] Change button danger background button color to `rgb(177, 30, 66)` when hovered (aka lightened up, to be similar behavior with other buttons on dark mode)
- [x] Crumb, on the last item (the active one), should have the same text color as the body; and other/previous links should be: `rgb(160, 160, 160)`.

  image
- [x] Pagination, for non active page, color is: `rgb(160, 160, 160)`, background and border color is: `rgb(55, 65, 81)`. 

  image

  On active: background: `rgb(97, 109, 128)`, border: `1px solid rgb(143, 143, 177)`, color: `#fff`. On disabled button, background: `rgb(56, 58, 61)`, color: `rgb(160, 160, 160)`

  image
- [x] On calendar, today's date color and circle color is `rgb(115, 94, 214)`

  image
- [x] Change all story "Docs" to have body and text color like all the other story (aka dark mode's default background and text color). The code box (like the `Usage` box which has class `prismjs`) can still be white. The preview also should use background color same like real story's background. Props table also background should blend in nicely.

  Screenshot from 2026-04-08 17-31-32

  Screenshot from 2026-04-08 17-31-38

  
  Screenshot from 2026-04-08 17-33-07


  Screenshot from 2026-04-08 17-39-44
- [x] Make coloring on `Dialog.show()` and other .show() method follows the global theme mode (even after color switch, currently only on first load does the theme get loaded and work, but if we switch the theme/mode, the old theme is the only that's used)

Truly, without storybook, this task is impossible to check and to be done. Will be crazy difficult.

theme/index.ts
```diff
@@ -0,0 +1,864 @@
...
+  textColor?: string;
+}
+
+// action-capsule
+export interface ActionCapsuleThemeConfiguration {
+  activeBackgroundColor?: string;
+  textColor?: string;
+  borderColor?: string;
+  capsuleWrapperBoxShadow?: string;
+  capsuleWrapperMinHeight?: string;
```
adamnoto: curious why we need to customize height as well?

theme/index.ts
```diff
@@ -0,0 +1,864 @@
...
+  action?: {
+    hoverBackgroundColor?: string;
+    activeBackgroundColor?: string;
+    focusRingColor?: string;
+    disabledOpacity?: number;
+  };
+}
+
+// boxbar.tsx
+export interface BoxbarThemeConfiguration extends BodyThemeConfiguration {
```
adamnoto: please define `BareColoringConfiguration` which includes `backgroundColor` and `textColor`; extending from `BodyThemeConfiguration` is scary... please do this for other configuration that extends the `Body` config, I think they don't need to do that.

theme/index.ts
```diff
@@ -0,0 +1,864 @@
...
+
+  rangeDateColor?: string;
+  highlightedDateTextColor?: string;
+  hightlightDateColor?: string;
+
+  boxShadow?: string;
+}
+
+// capsule.tsx
+export interface CapsuleThemeConfiguration {
```
adamnoto: extends `BareColoringConfiguration` and for anything that has background and textcolor let's make them extends the `BareColoringConfiguration`

theme/index.ts
```diff
@@ -0,0 +1,864 @@
...
+export interface KeynoteThemeConfiguration {
+  keyColor?: string;
+  valueColor?: string;
+}
+
+// list.tsx
+export interface ListThemeConfiguration {
+  backgroundColor?: string;
+  textColor?: string;
+  hoverBackgroundColor?: string;
```
adamnoto, line 350: let's create HoverableBareColoringConfiguration which extends BareColoringConfiguration but added `hoverBackgroundColor` and `hoverTextColor` and make `ListThemeConfiguration` or any config which has: `backgroundColor`, `textColor`, `hoverBackgroundColor` and `hoverTextColor` to extend from this `HoverableBareColoringConfiguration`

theme/index.ts
```diff
@@ -0,0 +1,864 @@
...
+
+  disabledBorderColor?: string;
+  disabledTextColor?: string;
+
+  placeholderColor?: string;
+  boxShadow?: string;
+}
+
+// phonebox.tsx
+export interface PhoneboxThemeConfiguration {
```
adamnoto: sorry, I think let's just name those configuration `Config` ie `PhoneboxThemeConfig`, `PinboxThemeConfig` having it `Configuration` is too wordy

theme/index.ts
```diff
@@ -0,0 +1,864 @@
...
+  starEmptyColor?: string;
+  starBorderColor?: string;
+
+  labelTextColor?: string;
+
+  disabledStarColor?: string;
+  disabledLabelColor?: string;
+
+  hoverStarColor?: string;
+  sizeMap?: { sm: number; md: number; lg: number };
```
adamnoto: why the size changes based on theme? 🤔 if they didn't differ now, let's not create this

theme/provider.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return () => {
+    listeners = listeners.filter((l) => l !== cb);
+  };
+}
+
+export function notifyThemeChange() {
+  listeners.forEach((cb) => cb());
+}
+
+export function ThemeProvider({
```
adamnoto, line 60: please give comment on this

theme/themes.ts
```diff
@@ -0,0 +1,2507 @@
...
+    ...customTheme,
+  };
+}
+
+// action-capsule
+export function createActionCapsuleTheme(
+  body: BodyThemeConfiguration,
+  customTheme: Partial<ActionCapsuleThemeConfiguration> = {}
+): ActionCapsuleThemeConfiguration {
+  const defaultTheme: ActionCapsuleThemeConfiguration = {
```
adamnoto: please extract all these kind of const to `theme/constants.ts` and name it like this:

`BASE_<COMPONENT_OR_ANY_OF_ITS_PART>_THEME_CONFIG` so:

- `createActionCapsuleTheme` -> `BASE_ACTION_CAPSULE_THEME_CONFIG`
- `createActionButtonTheme` -> `BASE_ACTION_BUTTON_THEME_CONFIG`
- `createAvatarTheme` -> `BASE_AVATAR_THEME_CONFIG`

I know they're actually for white/light mode, but it's okay, let's take "light" as the base for when we define theme config.

Make all those constants exportable.

theme/themes.ts
```diff
@@ -0,0 +1,2507 @@
...
+    activeBackgroundColor: "rgb(226, 224, 224)",
+    textColor: body.textColor || "#343434",
```
adamnoto: please constantantize `"#343434"` too if you feel they're used a lot, ie: `BASE_LIGHT_TEXT_COLOR` or `BASE_DARK_TEXT_COLOR`

so, for constants that are not theme config, please indicate if that's for `LIGHT` or  `DARK` theme

make all these constants exportable

of course not every literal values need to be a const, but for things that you feel/you think appears a lot, please constantize

adamnoto (changes requested):
Code-wise change request for now, on top of style guidelines update request [yesterday](https://github.com/systatum/coneto/pull/333#pullrequestreview-4073441293).

theme/index.ts
alimnfl: I've remove it that...

theme/themes.ts
```diff
@@ -0,0 +1,2626 @@
...
+  dividerHierarchySelectedColor: "#485c7d",
+});
+
+const darkWindow = createWindowTheme(darkBody, {
+  backgroundColor: "#111827",
+  textColor: "#f9fafb",
+  dividerColor: "#374151",
+});
+
+const darkTheme: AppTheme = {
```
adamnoto: let's move this to `theme/mode/dark.ts` and the light one to `theme/mode/light.ts`

and then import them here :)

let's also move `createStatusbarTheme` and other `createXTheme` into `theme/mode/creator.ts` and I think we can create `theme/mode/index.ts` and import * from creator into this `index.ts`

theme/index.ts
alimnfl: done

theme/index.ts
alimnfl: done

theme/provider.tsx
alimnfl: done

theme/themes.ts
alimnfl: done

adamnoto (approved):
Image

PR #335, opened 2026-04-10
[Improvement] SYST-597: Standardizing interface and field names

alimnfl:
Description:
We have NavTabContentProps where it has actions of type SubMenuNavTaband subitems of type NavTabSubItemsContentProps.

Let's reevaluate our interface naming convention, throughout all of our components. Let's have this rule:

If the interface name is X, the props interface to X is called XProps. So NavTab -> NavTabProps.

If inside that XProps there is a field/item which need custom class, use this format: XFieldNameProps. For example, take a look below.

```tsx
export interface NavTabProps {
  tabs?: NavTabContentProps[]; // Must be: NavTabTabProps[]
  actions?: NavTabActionsProps[]; // Must be: NavTabActionProps[]
  active?: boolean; // it is OK
  styles?: NavTabStylesProps; // Must be NavTabStyleProps, or just NavTabStyles
}

// DO NOT use plural ie NavTabTabsProps[] but singular NavTabTabProps[]
// the Props is plural, makes sense because containing a lot of properties,
// but the Tab is singular, because it is describing a single tab, not applicable
// to all the teb. That's why it's an array.
```

Another bad example. If the actions are "sub menu" then subitems is not sub menu? Why is actions "sub menu" but subitems are not "sub menu" if anything, subitems should be sub menu. Anyway, yeah, you can see that the naming is confusing. The name of the field is good, but the name of the class is very abstract and confusing — you will not know them just by looking at it, without looking at its usage.

Image you are not the one developing Coneto. When you see/heard the class/type "SubMenuNavTab" do you honestly think that it would be about action? NO, NEVER. I wouldn't. Only that I need to see the field name actions then I know the type is for navtab actions. This means, the field name is clear, but the class name is NOT clear, not self-describing.

```tsx
export interface NavTabContentProps {
  actions?: SubMenuNavTab[]; // WHAAAAT? very NON-STANDARD :D text for example: TextboxActionsProps
  subItems?: NavTabSubItemsContentProps[]; // WAIT WHAT?
  hidden?: boolean;
}

// To fix: rename SubMenuNavTab to: NavTabAction[]
// To fix: rename NavTabSubItemsContentProps[] (OMG SO LONG)
//         and an inconsistent naming, to: NavTabSubItem[]
```

Styles interface should be called: XStyles, so for NavTab it should be NavTabStyles, no need Props.

If the field has the word list or array, for example subMenuList, then the name of the interface is still singular, XSubMenuProps but the use is with []. X is the name of the component (ie: TipMenu).

```tsx
export interface TipMenuProps {
  subMenuList?: TipMenuSubMenuProps[];
}

// why? because List is already represented with the []
// let's say you want to define numberArray:

let numberArray: number[]

// right? NOT:

let numberArray: NumberArray[] // OR:
let numberArray: NumberArrays[] // BUT THIS IS CORRECT TOO:
let numberArray: NumberArray // however, why not just number[] ?

// as you can see, the "list" or "array" in the name
// is enough to represent with just []
```

Allow a little inconsistency, ie icon is using FigureProps not IconProps, that's ok. But most of the time, follow the ruling above. If not sure, let's discuss. IF YOU MAKE AN EXCEPTION/SKIP, please discuss. If not sure please discuss, I might miss not writing some rule.

Type like TipMenuItemVariantType must follow the same naming convention/technique above: XFieldName (without Type or Props).

```tsx
export type TipMenuItemVariantType = "sm" | "md";

export interface TipMenuProps {
  variant?: TipMenuItemVariantType;
}
```

For that, please rename TipMenuItemVariantType to: TipMenuVariant is clear, succinct, and enough. Also, please create the enum, so, the correct fix to that is like this:

```tsx
export const TipMenuVariant = {
  Small: "sm",
  Medium: "md",
}

export type TipMenuVariant =
  (typeof TipMenuVariant)[keyof typeof TipMenuVariant]

// by doing this, user can do TipMenuVariant.Small, or also
// just put "sm" as the value directly will be just as valid
```

Source:
[[Improvement] SYST-597: Standardizing interface and field names](https://linear.app/systatum/issue/SYST-597/standardizing-interface-and-field-names)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/action-button.tsx
+++ b/components/action-button.tsx
@@ -1,15 +1,17 @@
 import { css, CSSProp } from "styled-components";
-import { Button, ButtonVariants, SubMenuButtonProps } from "./button";
+import { Button, ButtonVariants, ButtonSubMenu } from "./button";
 import { ReactNode } from "react";
 import { FigureProps } from "./figure";
 import { useTheme } from "./../theme/provider";
+
+export type ActionButtonSubMenu = ButtonSubMenu;
 
 export interface ActionButtonProps {
   caption?: string;
   icon?: FigureProps;
   onClick?: () => void;
-  styles?: ActionButtonStylesProps;
-  subMenu?: (props: SubMenuButtonProps) => ReactNode;
+  styles?: ActionButtonStyles;
+  subMenu?: (props: ActionButtonSubMenu) => ReactNode;
   disabled?: boolean;
   showSubMenuOn?: "caret" | "self";
   variant?: ButtonVariants["variant"];
@@ -18,7 +20,7 @@ export interface ActionButtonProps {
   hidden?: boolean;
 }
 
-export interface ActionButtonStylesProps {
+export interface ActionButtonStyles {
   self?: CSSProp;
   toggleStyle?: CSSProp;
   dividerStyle?: CSSProp;
--- a/components/action-button.tsx
+++ b/components/action-button.tsx
@@ -3,14 +3,14 @@ import {
   Button,
   ButtonVariants,
   ButtonSubMenu,
-  ButtonShowSubMenuOn,
+  ButtonShowSubMenuPosition,
 } from "./button";
 import { ReactNode } from "react";
 import { FigureProps } from "./figure";
 import { useTheme } from "./../theme/provider";
 
 export type ActionButtonSubMenu = ButtonSubMenu;
-export type ActionButtonShowSubMenuOn = ButtonShowSubMenuOn;
+export type ActionButtonShowSubMenuPosition = ButtonShowSubMenuPosition;
 
 export interface ActionButtonProps {
   caption?: string;
@@ -19,7 +19,7 @@ export interface ActionButtonProps {
   styles?: ActionButtonStyles;
   subMenu?: (props: ActionButtonSubMenu) => ReactNode;
   disabled?: boolean;
-  showSubMenuOn?: ActionButtonShowSubMenuOn;
+  showSubMenuOn?: ActionButtonShowSubMenuPosition;
   variant?: ButtonVariants["variant"];
   className?: string;
   pressed?: boolean;
--- a/components/action-button.tsx
+++ b/components/action-button.tsx
@@ -1,10 +1,16 @@
 import { css, CSSProp } from "styled-components";
-import { Button, ButtonVariants, ButtonSubMenu } from "./button";
+import {
+  Button,
+  ButtonVariants,
+  ButtonSubMenu,
+  ButtonShowSubMenuOn,
+} from "./button";
 import { ReactNode } from "react";
 import { FigureProps } from "./figure";
 import { useTheme } from "./../theme/provider";
 
 export type ActionButtonSubMenu = ButtonSubMenu;
+export type ActionButtonShowSubMenuOn = ButtonShowSubMenuOn;
 
 export interface ActionButtonProps {
   caption?: string;
@@ -13,7 +19,7 @@ export interface ActionButtonProps {
   styles?: ActionButtonStyles;
   subMenu?: (props: ActionButtonSubMenu) => ReactNode;
   disabled?: boolean;
-  showSubMenuOn?: "caret" | "self";
+  showSubMenuOn?: ActionButtonShowSubMenuOn;
   variant?: ButtonVariants["variant"];
   className?: string;
   pressed?: boolean;
--- a/components/button.tsx
+++ b/components/button.tsx
@@ -70,11 +70,9 @@ export const ButtonDisplayLabel = {
 export type ButtonDisplayLabel =
   (typeof ButtonDisplayLabel)[keyof typeof ButtonDisplayLabel];
 
-export type ButtonTipMenu = TipMenuItemProps;
-
 export interface ButtonSubMenu {
   list?: (
-    subMenuList: ButtonTipMenu[],
+    subMenuList: TipMenuItemProps[],
     withFilter?: { withFilter?: boolean }
   ) => React.ReactNode;
   show?: (
@@ -84,17 +82,15 @@ export interface ButtonSubMenu {
   render?: (children?: React.ReactNode) => React.ReactNode;
 }
 
-export type ButtonTipMenuSize = TipMenuVariant;
-
 export type ButtonProps = Omit<React.ComponentProps<"button">, "style"> &
   ButtonVariants & {
     isLoading?: boolean;
     subMenu?: (props: ButtonSubMenu) => React.ReactNode;
     openedIcon?: FigureProps["image"];
     closedIcon?: FigureProps["image"];
-    tipMenuSize?: ButtonTipMenuSize;
+    tipMenuSize?: TipMenuVariant;
     safeAreaAriaLabels?: string[];
-    dialogPlacement?: ButtonDialogPlacement;
+    dialogPlacement?: DialogPlacement;
     onOpen?: (prop: boolean) => void;
     open?: boolean;
     styles?: ButtonStyles;
@@ -106,8 +102,6 @@ export type ButtonProps = Omit<React.ComponentProps<"button">, "style"> &
     showSubMenuOn?: ButtonShowSubMenuPosition;
     displayLabel?: ButtonDisplayLabel;
   };
-
-export type ButtonDialogPlacement = DialogPlacement;
 
 export interface ButtonStyles {
   dropdownStyle?: CSSProp | ((placement: Placement) => CSSProp);
--- a/components/button.tsx
+++ b/components/button.tsx
@@ -54,13 +54,13 @@ export type ButtonVariants = {
   size?: ButtonSize;
 };
 
-export const ButtonShowSubMenuOn = {
+export const ButtonShowSubMenuPosition = {
   Caret: "caret",
   Self: "self",
 } as const;
 
-export type ButtonShowSubMenuOn =
-  (typeof ButtonShowSubMenuOn)[keyof typeof ButtonShowSubMenuOn];
+export type ButtonShowSubMenuPosition =
+  (typeof ButtonShowSubMenuPosition)[keyof typeof ButtonShowSubMenuPosition];
 
 export const ButtonDisplayLabel = {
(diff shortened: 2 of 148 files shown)


components/tooltip.tsx
```diff
@@ -25,11 +25,19 @@ import { TooltipThemeConfig } from "./../theme";
...
+
+export type TooltipDialogOn =
+  (typeof TooltipDialogOn)[keyof typeof TooltipDialogOn];
+
 export type TooltipProps = {
   dialog: ReactNode;
   children: ReactNode;
-  showDialogOn?: "hover" | "click";
-  hideDialogOn?: "hover" | "click";
+  showDialogOn?: TooltipDialogOn;
```
adamnoto: I see, this one is a bit tricky and weird when we follow this pattern. If it ends with `On` let's change to `Position` -> `TooltipDialogPosition`

components/action-button.tsx
```diff
@@ -1,24 +1,32 @@
...
+  ButtonVariants,
+  ButtonSubMenu,
+  ButtonShowSubMenuOn,
+} from "./button";
 import { ReactNode } from "react";
 import { FigureProps } from "./figure";
 import { useTheme } from "./../theme/provider";
 
+export type ActionButtonSubMenu = ButtonSubMenu;
+export type ActionButtonShowSubMenuOn = ButtonShowSubMenuOn;
```
adamnoto: Please don't end with `On` but `Position` for something like this

adamnoto (changes requested):

components/action-button.tsx
alimnfl: done

components/tooltip.tsx
alimnfl: done

components/button.tsx
```diff
@@ -22,26 +22,59 @@ import { Figure, FigureProps } from "./figure";
...
+  OutlineSuccess: "outline-success",
+  OutlinePrimary: "outline-primary",
+  OutlineDanger: "outline-danger",
+} as const;
+
+export type ButtonVariant = (typeof ButtonVariant)[keyof typeof ButtonVariant];
+
+export const ButtonSize = {
+  Icon: "icon",
+  Xs: "xs",
```
adamnoto: let's give clear name: `ExtraSmall`, `Small`, `Medium`, and `Large` instead, but the mapped value can still use shorthand:

```ts
export const ButtonSize = {
  Icon: "icon",
  ExtraSmall: "xs",
  ....
}
```

components/button.tsx
```diff
@@ -51,28 +84,34 @@ export interface SubMenuButtonProps {
...
     pressed?: boolean;
     activeBackgroundColor?: string;
     hoverBackgroundColor?: string;
-    displayLabel?: "flex" | "ellipsis";
+    showSubMenuOn?: ButtonShowSubMenuPosition;
+    displayLabel?: ButtonDisplayLabel;
   };
 
-export interface ButtonStylesProps {
+export type ButtonIcon = FigureProps;
```
adamnoto: no need to do like this as I think it'd only make our type system convoluted for not much benefit; it's much clearer that we show/use the original type ...if we Ctrl + F "FigureProps" we will be able to see wherever it's used; basically a whole lot more benefit are there if we don't do type aliasing.

adamnoto (changes requested):

components/button.tsx
alimnfl: okay, I made this because on the task we need to make with similar case... but I'll change

adamnoto: hm, I just checked the doc but wasn't sure which point is this referring to. this one?

> Allow a little inconsistency, ie icon is using FigureProps not IconProps, that's ok. But most of the time, follow the ruling above. If not sure, let's discuss. IF YOU MAKE AN EXCEPTION/SKIP, please discuss. If not sure please discuss, I might miss not writing some rule.

components/button.tsx
alimnfl: done

components/button.tsx
alimnfl:
```tsx
export interface NavTabContentProps {
  actions?: SubMenuNavTab[]; // WHAAAAT? very NON-STANDARD :D text for example: TextboxActionsProps
  subItems?: NavTabSubItemsContentProps[]; // WAIT WHAT?
  hidden?: boolean;
}
```

this one, so I just naming by the component, because you said must to related like `NavTabAction`, so that name just to be like this, but yeah still need to discuss as you said at your quote

components/card.tsx
```diff
@@ -12,39 +12,63 @@ import { AnimatePresence, motion } from "framer-motion";
 import { useTheme } from "./../theme/provider";
 import { CardThemeConfig } from "./../theme";
 
+export const CardShadow = {
+  None: "none",
+  Sm: "sm",
```
adamnoto: `Small`, `Large` etc

components/datebox.tsx
```diff
@@ -1,22 +1,24 @@
...
-  SelectabilityModeState,
+  CalendarSelectabilityMode,
 } from "./calendar";
 import styled, { css, CSSProp } from "styled-components";
 import { forwardRef, ReactNode } from "react";
 import { FieldLaneProps } from "./field-lane";
 import { StatefulForm } from "./stateful-form";
 import { useTheme } from "./../theme/provider";
 
+type DateboxSelectabilityMode = CalendarSelectabilityMode;
```
adamnoto: no need type aliasing, can just use `CalendarSelectibilityMode`

components/dialog.tsx
```diff
@@ -44,16 +44,18 @@ export interface DialogProps {
...
   closable?: boolean;
-  styles?: DialogStylesProps;
+  styles?: DialogStyles;
   onClick?: (args: { id: string; closeDialog: () => void }) => void;
-  buttons?: DialogButtonProps[];
+  buttons?: DialogButton[];
   title?: ReactNode;
   subtitle?: ReactNode;
-  icon?: FigureProps;
+  icon?: DialogIcon;
```
adamnoto: please please always use FigureProps as it is, no need type aliasing

components/document-viewer.tsx
```diff
@@ -30,38 +30,51 @@ export type DocumentSource = (builder: {
   ) => { type: "encoded"; src: string };
 }) => ResolvedSource;
 
+export const DocumentViewerInitialZoom = {
```
adamnoto: hehe, this one is totally not necessary

can simplify to the old style

reasoning: user will never do like DocumentViewerInitialZoom.75, right? so we do this kind of style only if user can refer the value through its key using dot nation

adamnoto (changes requested):

components/card.tsx
alimnfl: done

components/document-viewer.tsx
alimnfl: done

components/dialog.tsx
alimnfl: done

components/datebox.tsx
alimnfl: done

components/button.tsx
```diff
@@ -54,25 +85,27 @@ export interface SubMenuButtonProps {
...
-    styles?: ButtonStylesProps;
+    styles?: ButtonStyles;
     anchorRef?: React.RefObject<HTMLElement>;
     icon?: FigureProps;
     pressed?: boolean;
     activeBackgroundColor?: string;
     hoverBackgroundColor?: string;
-    displayLabel?: "flex" | "ellipsis";
+    showSubMenuOn?: ButtonShowSubMenuPosition;
+    displayLabel?: ButtonDisplayLabel;
```
adamnoto: what's this for?

components/stateful-form.tsx
```diff
@@ -137,8 +169,8 @@ export interface FormFieldProps {
   labelGap?: FieldLaneProps["labelGap"];
   labelWidth?: FieldLaneProps["labelWidth"];
   disabled?: boolean;
-  rowJustifyContent?: "center" | "start" | "end" | "between";
-  rowAlignItems?: "center" | "start" | "end" | "between";
+  rowJustifyContent?: FormFieldRowJustifyContent;
```
adamnoto: `rowJustifyPosition` please rename the class too

components/stateful-form.tsx
```diff
@@ -137,8 +169,8 @@ export interface FormFieldProps {
...
+  rowAlignItems?: FormFieldRowAlignItems;
```
adamnoto: `rowItemsAlignment` please rename the interface too

components/table.tsx
```diff
@@ -62,28 +64,30 @@ export interface TableProps {
...
   onNextPageRequested?: () => void;
   onPreviousPageRequested?: () => void;
   disablePreviousPageButton?: boolean;
   disableNextPageButton?: boolean;
-  labels?: TableLabelsProps;
-  sumRow?: SummaryRowProps[];
-  styles?: TableStylesProps;
-  searchbox?: SearchboxProps;
+  labels?: TableLabels;
+  sumRow?: TableSummaryRow[];
```
adamnoto: `TableSummaryRow` -> `TableSummaryRowColumn` so become: `TableSummaryRowColumn[]`

adamnoto (changes requested):

components/table.tsx
adamnoto: this is because tho name is sumRow, we are explaining each column in that row, naming it `TableSummaryRow` is misleading because the naming is as if there're multiple rows

components/button.tsx
alimnfl: done

components/stateful-form.tsx
alimnfl: done

components/stateful-form.tsx
alimnfl: done

components/table.tsx
alimnfl: done

adamnoto (approved):
Image

PR #338, opened 2026-04-14
[Bug] SYST-627: Fix combobox coloring bug

alimnfl:
Description:
This pull request fixes the `highlighting` of the `selected` item in the drawer when re-opened via the `selectbox` or `arrow`. It also ensures consistent coloring across interactions (clicking, pressing).  

Source:
[[Bug] SYST-627: Fix combobox coloring bug](https://linear.app/systatum/issue/SYST-627/fix-combobox-coloring-bug)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/combobox.tsx
+++ b/components/combobox.tsx
@@ -504,6 +504,9 @@ function ComboboxDrawer({
             setHasInteracted(false);
           }
 
+          requestAnimationFrame(() => {
+            setHighlightedIndex(null);
+          });
           onClick?.();
         }}
         onMouseMove={() => {
--- a/components/combobox.tsx
+++ b/components/combobox.tsx
@@ -755,25 +755,18 @@ const listItemRowStyle = ({
     }
   `}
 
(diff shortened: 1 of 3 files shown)


adamnoto (changes requested):
I think this hasn't been implemented?

> Also, when hovering on an already selected item, I think no need to change the background color.

alimnfl:
> I think this hasn't been implemented?
> 
> > Also, when hovering on an already selected item, I think no need to change the background color.

done, sorry was missed.

adamnoto (changes requested):
nice cool cool cool, but I think the test code is missing when hovering selected item, should not change the background color. can we add that test please?

alimnfl:
> nice cool cool cool, but I think the test code is missing when hovering selected item, should not change the background color. can we add that test please?

done

adamnoto (approved):
finnncncnc

PR #360, opened 2026-04-17
[Bug] SYST-640: Fix hook error on Statusbar

alimnfl:
Description:
This pull request fixes an issue with context initialization when used within certain functionalities in the `Button` component related.

Previously, the context (i.e: `useTheme()`) could be initialized in an incorrect location, leading to unexpected behavior. In React, context must be defined and consumed within the proper hierarchy, specifically, at the top level of the component tree where it is intended to be shared.

image

Source:
[[Bug] SYST-640: Fix hook error on Statusbar](https://linear.app/systatum/issue/SYST-640/fix-hook-error-on-statusbar)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/button.tsx
+++ b/components/button.tsx
@@ -20,7 +20,7 @@ import {
 } from "./../lib/floating-placement";
 import { Figure, FigureProps } from "./figure";
 import { useTheme } from "./../theme/provider";
-import { TipMenuContainerThemeConfig } from "./../theme";
+import { AppTheme, TipMenuContainerThemeConfig } from "./../theme";
 
 export const ButtonVariant = {
   Link: "link",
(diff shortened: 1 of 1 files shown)


components/button.tsx
```diff
@@ -298,6 +303,7 @@ function Button({
           }}
         >
           <Divider
+            $theme={buttonTheme}
```
adamnoto, line 307: why is this buttonTheme when this is a `Divider`

alimnfl: because the `<Divider />` should adapts for the variant, you said before they need color for maybe `blend`, or anything

this also allowing user to customize the color through our theme.

```tsx
export interface ButtonThemeConfig extends BodyThemeConfig {
  hoverBackgroundColor?: string;
  activeBackgroundColor?: string;
  textDecoration?: string;
  focusBackgroundColor?: string;
  dividerColor?: string;
}
```

adamnoto (approved):
Image

PR #364, opened 2026-04-20
[Bug] SYST-643: Move libphonenumber and other really needed dependencies to be real dependencies

alimnfl:
Description:
This pull request fixes the our library to avoid using external sources, so the component can run without requiring any additional dependencies.

Source:
[[Bug] SYST-643: Move libphonenumber and other really needed dependencies to be real dependencies](https://linear.app/systatum/issue/SYST-643/move-libphonenumber-and-other-really-needed-dependencies-to-be-real)

Tick what you have done:
[x] I have tested it myself


lib/phone/index.ts
```diff
@@ -0,0 +1,645 @@
+// ---- CountryCode ----
+// Extracted from libphonenumber-js types (MIT License)
```
adamnoto, line 2: let's change the folder name to `phone`

and please include the license term of libphonenumber inside our own LICENSE

adamnoto (changes requested):

lib/phone/index.ts
alimnfl: you can see again.

adamnoto (approved):
Image

PR #367, opened 2026-04-20
[Improvement] SYST-649: Rename `id` on onClick for all dialog to be `buttonId`

alimnfl:
Description:
This pull request improves the `Button.onClick` handler inside the `Dialog` by renaming the `id` parameter to `buttonId`, making the argument name clearer and more explicit.

Source:
[[Improvement] SYST-649: Rename `id` on onClick for all dialog to be `buttonId`](https://linear.app/systatum/issue/SYST-649/rename-id-on-onclick-for-all-dialog-to-be-buttonid)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/test/component/dialog.cy.tsx
+++ b/test/component/dialog.cy.tsx
@@ -15,7 +15,7 @@ describe("Dialog", () => {
   }
 
   context("with Dialog.show()", () => {
-    it("renders the modal dialog", () => {
+    beforeEach(() => {
       cy.mount(
         <Button
           onClick={() =>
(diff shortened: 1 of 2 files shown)


test/component/dialog.cy.tsx
```diff
@@ -41,6 +45,25 @@ describe("Dialog", () => {
       cy.findByText("Default Modal").should("exist");
       cy.findByText("This is a subtitle").should("exist");
     });
+
+    it("renders the modal dialog", () => {
```
adamnoto: never ever write a blank test spec

I don't get what is the point of writing a spec in this style, why not just moving the `beforeEach` block inside of `it`?

adamnoto (changes requested):

test/component/dialog.cy.tsx
alimnfl: so this is to make the test to `re-usable`, so it will be represented from the `beforeEach`, but if you don't like it, I'll move them

alimnfl: done

adamnoto (approved):
Image

PR #370, opened 2026-04-23
[Improvement] SYST-604: Add `actions` to separator

alimnfl:
Description:
This pull request introduces an `actions` prop to the `Separator`, which can be positioned on the left or right depending on the `textFloat` prop. The `actions` prop defining items with `caption`, `icon`, `onClick`, `hidden`, `styles`, and `alwaysShow`. 

The `alwaysShow` property controls visibility behavior: when set to `true` (by default is false), the action is always visible; when `false`, it oly appears on hover over the separator.

Interface:
```tsx
export interface SeparatorAction {
  caption?: string;
  icon: FigureProps;
  alwaysShow?: boolean;
  onClick?: () => void;
  hidden?: boolean;
  styles?: SeparatorActionStyles;
}
```

Source:
[[Improvement] SYST-604: Add `actions` to separator](https://linear.app/systatum/issue/SYST-604/add-actions-to-separator)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/separator.stories.tsx
+++ b/components/separator.stories.tsx
@@ -24,6 +24,49 @@ Separator is a visual divider component that separates content into sections wit
 - ↔️ **Text float**: Title can float left or right.
 - 🧱 **Custom depth**: Adjust distance of the title from the container edge.
 - 🎨 **Customizable styles**: Supports container, line, and title styling via CSS props.
+
+### ⚡ Actions
+
+The \`actions\` prop allows you to attach interactive elements (such as icon buttons) to the separator.
+
+These actions are typically displayed on the opposite side of the title and can be used for quick operations like edit, copy, or open menus.
+
+---
+
+#### 🧱 Action Structure
+Each item in \`actions\` follows this shape:
+
+- \`icon\` (**required**) → Icon configuration (FigureProps)
+- \`caption\` → Tooltip text
+- \`onClick\` → Click handler
+- \`alwaysShow\` → Control visibility behavior (default: true)
+- \`hidden\` → Hide action completely
+- \`styles\` → Custom styles for button, tooltip, and positioning
+
+---
+
+#### 📌 Example
+\`\`\`tsx
+const actions = [
+  {
+    icon: { image: RiEditLine },
+    caption: "Edit",
+    onClick: () => console.log("Edit clicked"),
+  },
+  {
+    icon: { image: RiDeleteBinLine },
+    caption: "Delete",
+    alwaysShow: false,
+  },
+];
+\`\`\`
+
+---
+
+#### 🎯 Behavior Notes
+- Actions are automatically positioned with spacing.
+- Position adjusts based on \`textFloat\` (left or right).
+- Hovering the separator reveals actions when \`alwaysShow\` is false.
         `,
       },
     },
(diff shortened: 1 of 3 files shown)


components/separator.stories.tsx
```diff
@@ -19,6 +24,49 @@ Separator is a visual divider component that separates content into sections wit
...
+
+### ⚡ Actions
+
+The \`actions\` prop allows you to attach interactive elements (such as icon buttons) to the separator.
+
+These actions are typically displayed on the opposite side of the title and can be used for quick operations like edit, copy, or open menus.
+
+---
+
+#### 🧱 Action Structure
```
adamnoto: remove this, and combine the content into the `### ⚡ Actions` section

components/separator.stories.tsx
```diff
@@ -75,3 +136,18 @@ export const RightSide: Story = {
...
   },
 };
+
+export const WithActions: Story = {
+  args: {
+    title: "Antrikan App Redesign",
+  },
+  render: (args) => {
+    const ACTIONS: SeparatorAction[] = [
+      { icon: { image: RiUserAddLine }, caption: "Invite" },
```
adamnoto, line 143: let's make this `alwaysShow: false` for variation

test/component/separator.cy.tsx
```diff
@@ -0,0 +1,119 @@
...
+      <Separator actions={ACTIONS} title="Antrikan App Redesign" {...props} />
+    );
+  }
+  context("actions", () => {
+    beforeEach(() => {
+      cy.mount(<ProductSeparator />);
+    });
+
+    context("with alwaysShows", () => {
+      context("when given true", () => {
```
adamnoto: no need `given`

test/component/separator.cy.tsx
```diff
@@ -0,0 +1,119 @@
...
+                  icon: { image: RiUserAddLine },
+                  caption: "Invite",
+                  onClick: () => console.log("this is invite"),
+                  alwaysShow: false,
+                },
+              ]}
+            />
+          );
+        });
+        it("should not shows the actions", () => {
```
adamnoto, line 65: add enter before this `it` spec

adamnoto (changes requested):

test/component/separator.cy.tsx
alimnfl: done

test/component/separator.cy.tsx
alimnfl: done

components/separator.stories.tsx
alimnfl: done

components/separator.stories.tsx
alimnfl: done

adamnoto (changes requested):
the actions should be shown when we hover -not- just on the actions area, but also when the cursor is entering any part of that separator

alimnfl:
> the actions should be shown when we hover -not- just on the actions area, but also when the cursor is entering any part of that separator

it's work actually, really, maybe the line in the example is too small?

adamnoto (approved):
Image

PR #375, opened 2026-04-27
[Improvement] SYST-587: Remove `controlled` props

alimnfl:
Description:
This pull request removes the `controlled` in the `Combobox` component, which relied on the `Pagination`, since the current page need synchronize on the `combobox` when have behavior through the `navigation` buttons. Because of this, the selected value must always stay synchronized with the current page state.

One approach to address this is by using the `key` prop to made the `Combobox` always check after the page change.

```tsx
<Combobox key={`${currentPage}`} />
```

This ensures the combobox always reflects the latest value by remounting the component.

Source:
[[Improvement] SYST-587: Remove `controlled` props](https://linear.app/systatum/issue/SYST-587/remove-controlled-props)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/combobox.stories.tsx
+++ b/components/combobox.stories.tsx
@@ -146,10 +146,6 @@ Each field accepts a \`CSSProp\` (styled-components compatible) and can be used 
       control: "text",
       description: "Name/id used for accessibility and aria-label.",
     },
-    controlled: {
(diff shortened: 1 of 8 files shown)


adamnoto (approved):
Image

nice perfect, simple and "just perfect"

PR #376, opened 2026-04-27
[Improvement] SYST-605: Standardize `actions` on all components

alimnfl:
Description:
This pull request refactors the `actions` prop across components to inherit a shared  `BaseAction` interface. It also introduces proper `actions` support in components such as `EmptySlate`, and updates the `Dialog` component by replacing the `buttons` with `actions`. Additionally, it ensures all affected components remain fully functional after changes.

Source:
[[Improvement] SYST-605: Standardize `actions` on all components](https://linear.app/systatum/issue/SYST-605/standardize-actions-on-all-components)

Tick what you have done:
[x] I have tested it myself


components/combobox.tsx
```diff
@@ -611,10 +608,14 @@ function ComboboxDrawer({
...
 
+              if (action.hidden) {
+                return;
+              }
+
               return (
                 <Fragment key={index}>
                   <List.Item
-                    id={`action-${index}`}
+                    id={`action-${action.id}`}
```
adamnoto: what if `id` not given? `id` should be optional; if `id` is not given, please use the `index`

components/field-lane.tsx
```diff
@@ -234,16 +234,17 @@ function FieldLane({
           return (
             <Button
               key={index}
+              id={action.id}
```
adamnoto, line 240: curious what if `id` not given? if `null` will it skip setting the `id`? ideally we skip setting the `id` if it's not given.

components/field-lane.tsx
```diff
@@ -234,16 +234,17 @@ function FieldLane({
...
               labelMode="flex"
               aria-label="action-icon"
               onMouseDown={(e) => e.preventDefault()}
               onClick={(e) => {
                 e.stopPropagation();
-                if ((!disabled || !props?.disabled) && props.onClick) {
-                  props.onClick(e);
+                if ((!disabled || !action?.disabled) && action.onClick) {
```
adamnoto: I think we can simplify this kind of logic?

```tsx
if (disabled || action?.disabled) return
action?.onClick(e)
```

that code is easier to understand.

Think about the logic that we have there. Your intention is not to call right, if either it's `disabled` or `action.disabled` is true.

So, if `disabled` is true, `!disabled` is false. Ok. But what if `action.disabled` is false? It's `true`! And `true || false` is `true` !! So this code doesn't reflect you. intention.

Always be careful always when using complex boolean. Try to simplify it.

---

Please take care of this kind of code in this PR or in the codebase (there are other places as well like this).

components/list.stories.tsx
```diff
@@ -2088,6 +2098,7 @@ export const Accordion: Story = {
   render: () => {
     const ACTIONS_GROUPS: ListGroupaction[] = [
       {
+        id: "refresh",
```
adamnoto: remove all these changes, `id` should be optional

adamnoto (changes requested):
Please be careful and don't change unnecessarily. Forcing user to define `id` in this case is honestly a bad design and a bad engineering decision. Please don't cause unnecessary burden/stress on the user.

components/field-lane.tsx
alimnfl: it wouldn't show anything actually, render as usual

components/list.stories.tsx
alimnfl: done

components/field-lane.tsx
alimnfl: done

adamnoto (approved):
Image

nice!

PR #377, opened 2026-04-28
[Improvement] SYST-659: Add default class for each component and allow defining id

alimnfl:
Description:
This pull request standardizes `className` handling across all component at the parent level through `coneto-{component}` and introduces a consistent `id` prop to simplify element identification.

Source:
[[Improvement] SYST-659: Add default class for each component and allow defining id](https://linear.app/systatum/issue/SYST-659/add-default-class-for-each-component-and-allow-defining-id)

Tick what you have done:
[x] I have tested it myself


adamnoto (approved):
Image

PR #385, opened 2026-05-13
[Bug] SYST-680: Do not restrict what can be the children of a table

alimnfl:
Description:
This pull request fixes the `table` when the `content` was wrapped inside a separated component

Use case:
```tsx
    const columns: TableColumn[] = [
      {
        id: "itemId",
        caption: "Item ID",
        sortable: true,
      },
      {
        id: "name",
        caption: "Name",
        sortable: true,
        width: "60%",
      },
    ];

    function TableSeparateRow({ test }: { test: boolean }) {
      return <Table.Row content={["02", "Test 123"]} />;
    }

    return (
      <Table columns={columns}>
        <TableSeparateRow test={true} /> // previously rendered nothing
      </Table>
    );
```

Previously, `<TableSeparateRow />` would render nothing because `Table` only recognized direct `Table.Row` or `Table.Row.Group` elements as valid children.

Fixed by adding `resolveChildren()` which recursively unwraps wrapper components before processing.

It also fix the coloring the subtitle on the table.

Source:
[[Bug] SYST-680: Do not restrict what can be the children of a table](https://linear.app/systatum/issue/SYST-680/do-not-restrict-what-can-be-the-children-of-a-table)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/table.tsx
+++ b/components/table.tsx
@@ -1002,7 +1002,7 @@ function TableRowGroup({
             <TableRowGroupSubtitle
               aria-label="table-row-group-subtitle"
               $style={styles?.subtitleStyle}
-              color={tableTheme?.rowSubtitleTextColor}
+              $color={tableTheme?.rowGroupSubtitleTextColor}
             >
               {subtitle}
             </TableRowGroupSubtitle>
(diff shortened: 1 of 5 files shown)


adamnoto (approved):
Image

PR #388, opened 2026-05-15
[Improvement] SYST-679: Remove from coneto: `console.log(action.className)`

alimnfl:
Description:
Let's make a script to scan all files before we deploy to npm, to scan for anything that is console. in that file; and if there's such a file, fail abruptly and never allow to publish/submit the package.

Source:
[[Improvement] SYST-679: Remove from coneto: `console.log(action.className);`](https://linear.app/systatum/issue/SYST-679/remove-from-coneto-consolelogactionclassname)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/nav-tab.tsx
+++ b/components/nav-tab.tsx
@@ -418,7 +418,6 @@ function NavTab({
             {actions
               .filter((action) => !action?.hidden)
               .map((action, index) => {
-                console.log(action.className);
                 return (
                   <ActionButton
                     key={index}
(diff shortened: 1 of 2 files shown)


adamnoto (changes requested):
Let's not assume. Let's fail is we see console.log and let the developer decide what to do, instead of automatically not including it.

alimnfl:
> Let's not assume. Let's fail is we see console.log and let the developer decide what to do, instead of automatically not including it.

done for this

if failed:

image

else (succeed):

image

adamnoto (approved):
Image

PR #389, opened 2026-05-15
[Mobile] SYST-625: Add `mobile` mode for `Time` component

alimnfl:
Description:
This pull request introduces the `Wheel` component to `Coneto`, enabling a time-picker wheel experience similar to native mobile pickers. It also adds a `mobile` prop to support different interaction behaviors in mobile mode, preparing the component for future React Native compatibility.

Source:
[[Mobile] SYST-625: Add `mobile` mode for `Time` component](https://linear.app/systatum/issue/SYST-625/add-mobile-mode-for-time-component)

Tick what you have done:
[x] I have tested it myself


components/wheel.tsx
```diff
@@ -0,0 +1,453 @@
...
+}));
+
+const seconds = minutes;
+
+const ampm = [
+  { value: "am", text: "AM" },
+  { value: "pm", text: "PM" },
+];
+
+Wheel.fullHourOptions = fullHours;
```
adamnoto: let's not associate them with `Wheel`, `Wheel` should be generalized enough; we'll use this with combobox as well later :)

so, let's define this at the user-side (define it at Timebox, but no need `Timebox.hourOptions`, just define it as normal constant variable outside of the `Timebox`). and so at the Wheel stories, we also define again these variables within the story itself.

components/wheel.stories.tsx
```diff
@@ -0,0 +1,141 @@
...
+      width: "80px",
+    },
+  ]}
+/>
+\`\`\`
+
+- Use the \`parts\` prop to define each wheel column and its available values.
+- Use \`values\` and \`onChange\` for controlled state management.
+- Includes built-in presets:
+  - \`Wheel.hourOptions\`
```
adamnoto: no need this, Wheel is general; we just happen to use it to show time, so don't associate this specificity to `Wheel`

adamnoto (changes requested):

adamnoto:
On light please add proper background color to the wheel background, don't use white, they look too blended if using white.

components/wheel.tsx
alimnfl: done

components/wheel.stories.tsx
alimnfl: done

alimnfl:
> On light please add proper background color to the wheel background, don't use white, they look too blended if using white.

I've add by this one: "#e2e2e2"

options:

`#efefef`:
Screenshot 2026-05-20 at 09 42 59

`#e2e2e2`
Screenshot 2026-05-20 at 09 44 01

`#dddddd`
Screenshot 2026-05-20 at 09 44 34

`#cccccc`
Screenshot 2026-05-20 at 09 45 38

adamnoto (approved):
Image

PR #392, opened 2026-05-20
[Mobile] SYST-612: Add `mobile` to PaperDialog

alimnfl:
Description:
This pull request introduces the `height`, `controls`, and `mobile` props to `PaperDialog` to support a dedicated mobile appearance and behavior.

When `mobile` is set to `true`, left and right opening directions should still work, but only affect the animation behavior. In mobile mode, the dialog should automatically use `100%` screen width by default. Height should remain customizable, so users can still define a custom dialog height when needed.

Detail:

* Define `height` to control the dialog height, with a default value of `100dvh` when not provided.
* Define `width` to control the dialog width, with a default value of `70vh` when not provided.
* Move the minimize/restore and close controls to the top area when `mobile` is enabled, since the current left/right indicator can feel a bit misleading on mobile layouts.
* Introduce a `controls` prop. By default (or when `null` / `undefined`), it should be:

  ```ts
  ["minimize", "close"]
  ```

  The `"minimize"` control should include both restore (`<`) and minimize (`>`) behaviors. This also allows mobile consumers to hide minimize or close controls if needed.
* Show a draggable indicator line at the top of the mobile dialog, allowing users to drag the dialog downward with a hand gesture to close it, similar to mobile bottom sheets or sidebars.

Other existing behaviors should still apply. For example, when `closable={false}`, the close button should be hidden and the dialog should not close when clicking outside of it.

Source:
[[Mobile] SYST-612: Add `mobile` to PaperDialog](https://linear.app/systatum/issue/SYST-612/add-mobile-to-paperdialog)

Tick what you have done:
[x] I have tested it myself


adamnoto (approved):
Image

PR #397, opened 2026-05-21
[Bug] SYST-672: Fix bug cannot set `id` to rich editor

alimnfl:
Description:
This pull request fixes the `id` in the `RichEditor` so it is properly applied to the parent element. It also ensures the `id` correctly forwarded to the underlying element. 

Source:
[[Bug] SYST-672: Fix bug cannot set `id` to rich editor](https://linear.app/systatum/issue/SYST-672/fix-bug-cannot-set-id-to-rich-editor)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/rich-editor.tsx
+++ b/components/rich-editor.tsx
@@ -152,10 +152,9 @@ export interface RichEditorToolbarButtonStyles {
   self?: CSSProp;
 }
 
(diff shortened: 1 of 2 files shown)


adamnoto (approved):
Image

PR #399, opened 2026-05-21
[Improvement] SYST-688: Add `mobile` on `combobox`

alimnfl:
Description:
This pull request introduces `mobile` mode on the `Combobox`, providing a bottom screen appearance similar to the Timebox (wheel), including `fade up` and `fade bottom` effects. It also ensures all related visual appearances are properly implemented in the component.

Source:
[[Improvement] SYST-688: Add `mobile` on `combobox`](https://linear.app/systatum/issue/SYST-688/add-mobile-on-combobox)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -1,7 +1,10 @@
 <head>
   <meta charset="UTF-8" />
   <title>Systatum Coneto React UI Library</title>
-  <meta name="viewport" content="width=device-width, initial-scale=1" />
+  <meta
+    name="viewport"
+    content="width=device-width, initial-scale=1, maximum-scale=1"
+  />
   <meta
     name="description"
     content="Deliver apps in half the time with Coneto — a complete React UI library by Systatum, built for speed and modern design."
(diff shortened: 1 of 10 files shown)


adamnoto (changes requested):
I think let's push other variations as well in the same "Mobile"  so checkbox (with title: "Selectable combobox") and then one with nested group (title: "With grouped options") --- should be enough

alimnfl:
> I think let's push other variations as well in the same "Mobile" so checkbox (with title: "Selectable combobox") and then one with nested group (title: "With grouped options") --- should be enough

done for this, I've add the appearance on `categorized mode`, check it again....

adamnoto (changes requested):
I think one more and we good to go:

When mobile, let's make it works like a `strict` (so setting strict to false or true has no meaning when mobile). So, whenever we click on a combobox, let's always show the options.

alimnfl:
> I think one more and we good to go:
> 
> When mobile, let's make it works like a `strict` (so setting strict to false or true has no meaning when mobile). So, whenever we click on a combobox, let's always show the options.

done for this.

adamnoto (approved):
Image

PR #406, opened 2026-05-28
[Mobile] SYST-698: Add `mobile` to button

alimnfl:
Description:
This pull request introduces a `mobile` mode for the `Button` component. It adds mobile-specific customization, including disabling hover effects and rendering the button slighly larger by adjusting properties such as `font-size` and `height`. 

Source:
[[Mobile] SYST-698: Add `mobile` to button](https://linear.app/systatum/issue/SYST-698/add-mobile-to-button)

Tick what you have done:
[x] I have double checked the functionality with the ticket in linear, or any other relevant discussion avenue
[x] I have created/updated any relevant test code
[x] I have tested it myself

PR #407, opened 2026-05-28
[Mobile] SYST-691: Improve Dialog mobile look and feel

alimnfl:
Description:
This pull request improves `PaperDialog` and `Dialog` components by preventing unintended text selection behavior inside both components. It also ensures that buttons inside `Dialog` do not display hover behavior on mobile interactions.

The button will make with separate task to this [pull request](https://github.com/systatum/coneto/pull/406)
 
Source:
[[Mobile] SYST-691: Improve Dialog mobile look and feel](https://linear.app/systatum/issue/SYST-691/improve-dialog-mobile-look-and-feel)

Tick what you have done:
[x] I have tested it myself


adamnoto (changes requested):
> It also ensures that buttons inside Dialog do not display hover behavior on mobile interactions.

This is not what the task wants tho:

> And when we tap (or press/thumb-press) on the button, make it as if user is hovering (because on mobile we can't use hover, so the color indication should change differently, ie when pressed).

So the task description is exactly opposite to what the task says 🥲 . Basically, let's make those buttons work like `mobile` mode (we have mobile mode, just let's set those button `mobile` too when dialog mode is mobile)

alimnfl:
> > It also ensures that buttons inside Dialog do not display hover behavior on mobile interactions.
> 
> This is not what the task wants tho:
> 
> > And when we tap (or press/thumb-press) on the button, make it as if user is hovering (because on mobile we can't use hover, so the color indication should change differently, ie when pressed).
> 
> So the task description is exactly opposite to what the task says 🥲 . Basically, let's make those buttons work like `mobile` mode (we have mobile mode, just let's set those button `mobile` too when dialog mode is mobile)

done

adamnoto (approved):
Image

PR #410, opened 2026-05-29
[Improvement] SYST-697: Don't regenerate feed if nothing changes

alimnfl:
Description:
This pull request improves the `feed` to not shows if not any changes inside of the `component/stories` with this terms:

1. Define a new xmlns at the rss parent declaration, like this below. For our case, maybe name it xmlns:coneto
```
<?xml version="1.0"?>
<rss version="2.0" xmlns:w="http://tempuri.org">
    <channel>
        ...
        <item>
            <title>Item title</title>
            <description>...</description>
            <w:temperature>...</w:temperature>
            <w:windspeed>...</w:windspeed>
        </item>
    </channel>
</rss>
```

2. Then for each item, add coneto:item-digest which is an MD5 digest of the content:encoded part of the item.

3. Whenever we attempt to generate the feed.xml:
     a. Generate the item link
     b. Find item of the same link in existing feed.xml
          i. If not found, add the item directly
          ii. If found:
              i. Check digest, if digest is different, replace the item with the new content, and so update the new digest and pubDate
              ii. Skip, no update needed. Don't change anything.

Source:
[[Improvement] SYST-697: Don't regenerate feed if nothing changes](https://linear.app/systatum/issue/SYST-697/dont-regenerate-feed-if-nothing-changes)

Tick what you have done:
[ ] I have tested it myself

### DIFF
--- a/public/feed.xml
+++ b/public/feed.xml
@@ -2,7 +2,8 @@
 <rss version="2.0"
   xmlns:dc="http://purl.org/dc/elements/1.1/"
   xmlns:atom="http://www.w3.org/2005/Atom"
-  xmlns:content="http://purl.org/rss/1.0/modules/content/">
+  xmlns:content="http://purl.org/rss/1.0/modules/content/"
+  xmlns:coneto="https://coneto.systatum.com/rss/extensions">
   <channel>
     <title>Systatum Coneto React UI Library</title>
     <link>https://coneto.systatum.com</link>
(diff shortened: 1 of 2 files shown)


adamnoto (changes requested):
Please run the generate-rss script and commit the changes for the feed

adamnoto (approved):
finnncncnc

PR #413, opened 2026-06-03
[Mobile] SYST-707: Make it easy to set how large is the `drawer` on `combobox mobile` mode

alimnfl:
Description:
This pull request enhances the `mobile` prop of `Combobox` to support both a `boolean` or `object` value. Previously, the prop only accepted a boolean.

This approach provides a simple and scalable API, allowing future mobile-specific customizations to be introduced through a simple prop without adding additional configuration props.

```tsx
<Combobox mobile={{ name: "A" }} />
```

The existing boolean usage remains fully supported:

```tsx
<Combobox mobile />
```

It also ensures this approach works correctly, and introduces a `drawerHeight` inside of the `mobile` props.

Source:
[[Mobile] SYST-707: Make it easy to set how large is the drawer on combobox mobile mode](https://linear.app/systatum/issue/SYST-707/make-it-easy-to-set-how-large-is-the-drawer-on-combobox-mobile-mode)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/combobox.tsx
+++ b/components/combobox.tsx
@@ -56,7 +56,11 @@ interface BaseComboboxProps {
   navigableOptions?: SelectboxOption[];
   isLoading?: boolean;
   labels?: ComboboxLabelsProps;
-  mobile?: boolean;
+  mobile?: boolean | ComboboxMobile;
+}
+
+export interface ComboboxMobile {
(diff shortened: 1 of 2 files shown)


adamnoto (approved):
Image

PR #417, opened 2026-06-08
[Research] SYST-702: Investigate how to extend RichEditor

alimnfl:
Description:
This PR introduces a new `tokenRenderers` prop for `RichEditor`, enabling custom inline token rendering directly from `Markdown` content.

`tokenRenderers` provides a flexible way to define custom token syntax and render it as React components while preserving normal editor behavior.

```tsx
<RichEditor
  tokenRenderers={{
    "<{[": {
      endToken: "]}>",
      render: (word) => {
        const parsed = JSON.parse(`[${word}]`);
        const [surah, ayah] = parsed;

        return <Badge caption={`${surah}:${ayah}`} />;
      },
    },
  }}
/>
```

Markdown:
This is a verse reference <{["Q", "17:32"]}> inside text.

Rendered output:
This is a verse reference [Badge: Q:17:32] inside text.

Weakness for now:
Can't render the `<div/>` element properly, if inside of paragraph.

Source:
[[Research] SYST-702: Investigate how to extend RichEditor](https://linear.app/systatum/issue/SYST-702/investigate-how-to-extend-richeditor)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/badge.tsx
+++ b/components/badge.tsx
@@ -16,10 +16,8 @@ export const BadgeVariant = {
 
 export type BadgeVariant = (typeof BadgeVariant)[keyof typeof BadgeVariant];
 
-export interface BadgeProps extends Omit<
-  HTMLAttributes<HTMLDivElement>,
-  "style"
-> {
+export interface BadgeProps
+  extends Omit<HTMLAttributes<HTMLDivElement>, "style"> {
   id?: string;
   metadata?: Record<string, unknown>;
   variant?: BadgeVariant;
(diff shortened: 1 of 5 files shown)


adamnoto (changes requested):
Looks cool! I tried to delete right, the tag, then Ctrl+Z on it, so the tag appears again, but when I print, it doesn't appear:

Image

alimnfl:
> Looks cool! I tried to delete right, the tag, then Ctrl+Z on it, so the tag appears again, but when I print, it doesn't appear:
> 
> Image

done for this, right now I think was applied properly (after delete and undo).

adamnoto (approved):
Image

PR #421, opened 2026-06-10
[Feature] SYST-718: Add `<Flippable/>`

alimnfl:
Description:
This pull request introduces a `Flippable` component that enables smooth front-to-back card flipping animations.

### Props

* `children` — Content displayed on the front face.
* `back` — Content displayed on the back face.
* `width` — Sets the component width.
* `height` — Sets the component height.
* `styles` — Allows custom styling for the container, front face, and back face.
* `onFlip` — Callback triggered when the flip state changes.
* `className` — Additional CSS class names.
* `id` — HTML id attribute for the root element.
* `flipDuration` — Controls the flip animation duration.

### Features

* Supports customizable front and back content.
* Exposes imperative methods (`flip`, `unFlip`, and `toggle`) through refs.
* Integrates with the theme system for front and back face appearance.
* Provides configurable dimensions and animation timing.
* Supports custom styling for each face and the container.
* Automatically toggles between states when clicked.

Source:
[[Feature] SYST-718: Add `<Flippable/>`](https://linear.app/systatum/issue/SYST-718/add-flippable)

Tick what you have done:
[x] I have tested it myself


adamnoto (changes requested):
Rather than having `flipOnClick` let's have `onClick="flip"` (which is the default) or `"preventDefault"` (which does nothing), or user can provide a function:

```ts
onClick={({ preventDefault, flip }) => {
  if (a) preventDefault();
  else flip();
}}
```

alimnfl:
> Rather than having `flipOnClick` let's have `onClick="flip"` (which is the default) or `"preventDefault"` (which does nothing), or user can provide a function:
> 
> ```ts
> onClick={({ preventDefault, flip }) => {
>   if (a) preventDefault();
>   else flip();
> }}
> ```

Okie, I've change with this API, to enrich flexibility, maybe users want to use `stopPropagation`, or anything event available from `div` element.

```tsx
 onClick?: (props: {
    event?: React.MouseEvent<HTMLDivElement>;
    flip?: () => void;
    unFlip?: () => void;
    toggle?: () => void;
  }) => void;
```

components/flippable.stories.tsx
```diff
@@ -0,0 +1,478 @@
...
+          flexDirection: "column",
+          gap: 16,
+          alignItems: "center",
+        }}
+      >
+        <Flippable
+          ref={ref}
+          width={420}
+          height={350}
+          flipDuration={0.7}
```
adamnoto: I think `speed` is a better name, and is clear enough

adamnoto (changes requested):

components/flippable.stories.tsx
alimnfl: done

adamnoto (approved):
finnnnn

PR #424, opened 2026-06-11
[Bug] SYST-727: Remove height: 100% from `MessageBox`

alimnfl:
Description:
This pull request fixes an issue where `Messagebox` always inherited a height of `100%` from its parent container. As a result, if the parent had additional empty space (e.g. `200px`), the `Messagebox` would unnecessarily expand to fill it.

The `Messagebox` now uses `fit-content` by default, allowing its height to be determined by its content rather than the available space in the parent container.

Source:
[[Bug] SYST-727: Remove height: 100% from `MessageBox`](https://linear.app/systatum/issue/SYST-727/remove-height-100percent-from-messagebox)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/messagebox.tsx
+++ b/components/messagebox.tsx
@@ -70,6 +70,7 @@ function Messagebox({
   return (
     <Wrapper
       id={id}
+      aria-label="messagebox-container"
       className={applyClassName("messagebox", className)}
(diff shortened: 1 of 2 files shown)


adamnoto (approved):
Image

PR #428, opened 2026-06-15
[Bug] SYST-730: Apply `Styles` Correctly to `Title` Actions

alimnfl:
Description:
This pull request fixes an issue where the `styles` prop on the `Title` component was not applied correctly when defined in `containerStyle` or `self`.

The issue occurred because the `Title` component uses a different rendering mechanism from other components when used within a `ContextMenu`. While other components are automatically moved into a `TipMenu`, `Title` must remain outside the `TipMenu`, which requires additional configuration to ensure its styles are applied correctly. This change enables `Title` styles to work as expected in both `containerStyle` and `self`.

```tsx
<Title
  rightSection={[
    {
      type: "actions",
      actions: [
        {
          icon: {
            image: RiSettings5Line,
            color: TEXT_COLOR,
          },
          onClick: () => {
            onActionClicked(ContentType.Settings);
          },
          styles: {
            containerStyle: css`
              background: black;
            `,
            self: css`
              background: blue;
            `,
          },
        },
      ],
    },
  ]}
/>
```

Source:
[[Bug] SYST-730: Apply `Styles` Correctly to `Title` Actions](https://linear.app/systatum/issue/SYST-730/apply-styles-correctly-to-title-actions)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/test/component/title.cy.tsx
+++ b/test/component/title.cy.tsx
@@ -23,7 +23,7 @@ describe("Title", () => {
   });
 
   context("styles", () => {
-    context("title Actions (specific)", () => {
+    context("actions styles", () => {
       context("containerStyle", () => {
         context("when given border blue color", () => {
           it("renders the blue colors on the container", () => {
--- a/test/component/title.cy.tsx
+++ b/test/component/title.cy.tsx
@@ -1,4 +1,4 @@
-import { RiCloseFill } from "@remixicon/react";
+import { Ri24HoursFill, RiCloseFill, RiCloseLine } from "@remixicon/react";
 import { Title } from "./../../components/title";
 import { Button } from "./../../components/button";
 import { css } from "styled-components";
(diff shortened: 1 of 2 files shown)


test/component/title.cy.tsx
```diff
@@ -23,6 +23,94 @@ describe("Title", () => {
   });
 
   context("styles", () => {
+    context("title Actions (specific)", () => {
```
adamnoto: why is actions capitalized? why not just `on section actions` ?  curious also why the wording `(specific)` ?

adamnoto (changes requested):

test/component/title.cy.tsx
alimnfl: I mean specific styling on the `actions`, I'll changes this.

adamnoto (approved):
Image

PR #429, opened 2026-06-15
[Feature] SYST-720: Add `<Carousel>` on "Stage"

alimnfl:
Description:
This pull request introduces a `Carousel` component in the stage categor, enabling users to navigate between multiple pieces of content using swipe gestures, navigation arrows, keyboard controls, and pagination indicators.

`Carousel Props`.
* `length`: Specifies the total number of pages in the carousel. If not provided, the carousel automatically uses the number of children as the page count.
* `currentPage`: Controls the currently active page. When provided, the carousel operates in controlled mode and page changes must be handled externally.
* `onChange`: Callback invoked when the user requests navigation to the `next/prev` page.
* `initialPage`: Specifies the initial page when the carousel is used in uncontrolled mode. Defaults to `0`.
* `control`: Enables pagination controls. Pass `true` to use the default controls, or provide a `CarouselControl` object to customize their position.
* `styles`: Custom styles for the carousel container, content, controls, and navigation arrows.
* `children`: The carousel content. Each direct child is treated as a separate slide.
* `className`: Additional CSS class names applied to the carousel root element.
* `id`: HTML `id` attribute applied to the carousel root element.

Source:
[[Feature] SYST-720: Add `<Carousel>` on "Stage"](https://linear.app/systatum/issue/SYST-720/add-carousel-on-stage)

Tick what you have done:
[x] I have tested it myself


adamnoto (changes requested):
I think let's do this for control:

```ts
controller={{
  circle: Carousel.CircleControllerPosition.BottomCenter,,
  arrow: null
}}
```

In the case above, just have the circle controller, without arrow. Or the other way around:

```ts
controller={{
  circle: null,,
  arrow: Carousel.ArrowControllerPosition.CenterSide // so it's left and right at center side
}}
```

Possible value for arrow right now: `CenterSide` (only).

Or of course we can enable both arrow and circle controller.

If `control` is not defined, let's not assume: just don't render any control, but allow sliding left and/or right as usual.

Also let's add a way to style them both, ie: `arrowControllerStyle`, `circleControllerStyle`, and `circleControllerCurrentPageStyle`

adamnoto (changes requested):
Hm, this API is weird:

```ts
<Carousel controller autoHeight>
```

`controller` cannot be a boolean; it must be an object. so if `controller` is not defined, don't render the controller; aka you missed this point:

> If control is not defined, let's not assume: just don't render any control, but allow sliding left and/or right as usual.

🙂

alimnfl:
> `controller` cannot be a boolean; it must be an object. so if `controller` is not defined, don't render the controller; aka you missed this point:

done for this, sorry for the mistake :)

adamnoto (approved):
Image

PR #431, opened 2026-06-17
[Bug] SYST-738: Do not make status to be `absolute`

alimnfl:
Description:
This pull request updates the `Statusbar` component to use `position: absolute` by default, ensuring it remains consistently positioned at the bottom of the viewport while scrolling. The component continues to support both `fixed` and `absolute` positioning through the `position` prop.

If a bottom spacing/gap is required, use the `Statusbar` component itself rather than relying on external spacing, as the fixed positioning may otherwise cause content to be visually clipped at the bottom of the viewport.

Source:
[[Bug] SYST-738: Do not make status to be absolute](https://linear.app/systatum/issue/SYST-738/do-not-make-status-to-be-absolute)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/statusbar.stories.tsx
+++ b/components/statusbar.stories.tsx
@@ -12,6 +12,7 @@ import {
   RiUserFollowLine,
   RiWindow2Fill,
 } from "@remixicon/react";
+import { generateSentence } from "./../lib/text";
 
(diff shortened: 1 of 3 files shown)


adamnoto (approved):
finnncncnc

PR #432, opened 2026-06-18
[Improvement] SYST-740: Support `variant` for `row-actions` when less than `maxActionsBeforeCollapsing`

alimnfl:
Description:
This pull request fixes the `ContextMenu` so that when the number of actions is less than `maxActionsBeforeCollapsing`, it properly supports per-action variants. This behavior is applied consistently in components such as `Table`, `TreeList`, `List`, and `NavTab`. It also ensures that all four components render correctly.

Source:
[[Improvement] SYST-740: Support variant for row-actions when less than maxActionsBeforeCollapsing](https://linear.app/systatum/issue/SYST-740/support-variant-for-row-actions-when-less-than)

Tick what you have done:
[x] I have tested it myself


adamnoto (approved):
Image

PR #436, opened 2026-06-22
[Bug] SYST-750: Fix `z-index` on Toast

alimnfl:
Description:
This pull request fixes a toast `z-index` issue where the toast is not visible when rendered inside components such as `<PaperDialog />` or other elements with a higher `z-index` than the toast itself.

It also ensures that the `Toast` is correctly displayed when used inside `<PaperDialog />` and introduces a new `z-index` prop to enrich flexibility index for the `Toast` component.

Source:
[[Bug] SYST-750: Fix `z-index` on Toast](https://linear.app/systatum/issue/SYST-750/fix-z-index-on-toast)

Tick what you have done:
[x] I have tested it myself


adamnoto (approved):
Image

PR #437, opened 2026-06-22
[Improvement] SYST-600: Add `loose` property to `Table`

alimnfl:
Description:
This pull request introduces a new `loose` prop for the `Table` component, providing an enhanced layout and scrolling experience for tables with a large number of columns (typically more than 12).

The `loose` mode includes:

* Improved horizontal scrolling behavior for table content.
* Sticky positioning for the first column to maintain context while scrolling.
* Sticky positioning for `row-actions` on the right side of the table.
* Additional styling and customization options to better support wide-table use cases.

These enhancements improve usability and readability when working with large datasets while preserving access to key information and actions during horizontal scrolling.

Note:
In the `Table` case, scrollbar behavior may become problematic in the future because scrollbar widths vary across browsers and operating systems, making calculations inconsistent, and affected to `row actions` in `Table.Loose` feature.

For experiment, at first, I used an npm package for the scrollbar. It worked well, but the implementation is actually quite simple, so we can build it ourselves instead of adding another dependency.

The idea is simple: add a track for `scrollX` at the bottom and a track for `scrollY` on the right side. Then add a draggable thumb to handle scrolling. We can also support theming to make the scrollbar easy to customize and keep consistent with the rest of the components. 

To address this, we introduced the `Scrollbar` component inside the `Table`, ensuring consistent behavior and reducing browser- and OS-specific discrepancies. 

Source:
[[Improvement] SYST-600: Add `loose` property to `Table`](https://linear.app/systatum/issue/SYST-600/add-loose-property-to-table)

Tick what you have done:
[x] I have tested it myself


adamnoto (changes requested):
This is in my macOS, still having problem
 
Image

also, when on right most, can we revert the "inert" style, intinya, yang "mendelep" atau masuk harusnya bukan si action; dalam arti lain, row si action hari tara lebih "raised" kalo sekarang si row action terasa lebih "rendah" ..

yang lebih "rendah" konsisten harus data yang ditengah; kanan kiri harus terasa lebih naik

alimnfl:
> This is in my macOS, still having problem
> 
> Image
> also, when on right most, can we revert the "inert" style, intinya, yang "mendelep" atau masuk harusnya bukan si action; dalam arti lain, row si action hari tara lebih "raised" kalo sekarang si row action terasa lebih "rendah" ..
> 
> yang lebih "rendah" konsisten harus data yang ditengah; kanan kiri harus terasa lebih naik

done

test/component/table.cy.tsx
```diff
@@ -187,6 +192,470 @@ describe("Table", () => {
...
+              .then(($el) => {
+                const after = window.getComputedStyle($el[0], "::after");
+
+                expect(after.backgroundImage).to.not.equal(
+                  "linear-gradient(to right, rgba(0, 0, 0, 0.08), rgba(0, 0, 0, 0))"
+                );
+              });
+          });
+
+          context("when scrolling to the right", () => {});
```
adamnoto: is this test missing? during dev, please make sure to fill it with `expect(true).toBe(false)` or things like that to help you catch "test that I gonna write for later"

adamnoto (changes requested):

test/component/table.cy.tsx
alimnfl: it was tested actually, just missed to add inside `it` into `context`, for expect the value it was properly, always using `to.equal` or `to.not.equal`

```tsx
cy.get(".coneto-button")
  .eq(3)
  .then(($el) => {
    const before = window.getComputedStyle($el[0], "::before");

    expect(before.backgroundImage).to.equal("none");
  });
```

adamnoto (approved):
Image

PR #441, opened 2026-06-24
[Bug] SYST-756: `Toolbar.Menu` was hidden if opened inside of overflow `scroll` or `auto`

alimnfl:
Description:
Previously, the `Toolbar` did not render correctly when used inside `Table.subMenuList`, particularly when the parent container had `overflow: auto` or `overflow: scroll`. As a result, the `TipMenu` could be clipped or displayed improperly when opened.

This pull request fixes the issue by rendering the `TipMenu` through a React Portal and mounting it directly to `document.body`. This ensures the dropdown is displayed correctly regardless of the overflow behavior of its parent containers.

Source:
[[Bug] SYST-756: `Toolbar.Menu` was hidden if opened inside of overflow `scroll` or `auto`](https://linear.app/systatum/issue/SYST-756/toolbarmenu-was-hidden-if-opened-inside-of-overflow-scroll-or-auto)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/toolbar.tsx
+++ b/components/toolbar.tsx
@@ -23,6 +23,7 @@ import { Figure, FigureProps } from "./figure";
 import { useTheme } from "./../theme/provider";
 import { ToolbarThemeConfig } from "./../theme";
 import { applyClassName } from "./../constants/classname";
+import { createPortal } from "react-dom";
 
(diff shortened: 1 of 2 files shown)


adamnoto (approved):
Image

PR #442, opened 2026-06-24
[Improvement] SYST-755: Extract `drawerHeight` out of `mobile`

alimnfl:
Description:
This pull request simplifies and clarifies the API by making `drawerHeight` prop available for both desktop and mobile layouts, rather than limiting it to the `mobile` configuration. As a result, the `mobile` prop is now a simple boolean instead of an object.

It also ensures that `drawerHeight` is applied correctly and consistently across all supported layouts.

Source:
[[Improvement] SYST-755: Extract `drawerHeight` out of `mobile`](https://linear.app/systatum/issue/SYST-755/extract-drawerheight-out-of-mobile)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/combobox.tsx
+++ b/components/combobox.tsx
@@ -56,10 +56,7 @@ interface BaseComboboxProps {
   navigableOptions?: SelectboxOption[];
   isLoading?: boolean;
   labels?: ComboboxLabelsProps;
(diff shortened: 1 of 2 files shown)


adamnoto (approved):
Image

PR #443, opened 2026-06-24
[Bug] SYST-754: Fix weird focus coloring on chips

alimnfl:
Description:
This pull request fixes an issue when clicking a `Chips.Item` caused the `textbox` focus to flicker. It also ensures that focus remains on the `textbox` after an option is selected, without triggering an additional focus event on the `Chips.Item`.

Source:
[[Bug] SYST-754: Fix weird focus coloring on chips](https://linear.app/systatum/issue/SYST-754/fix-weird-focus-coloring-on-chips)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/chips.tsx
+++ b/components/chips.tsx
@@ -438,7 +438,6 @@ function ChipsDrawer({
                         !isClicked && (
                           <Divider $theme={chipsTheme} aria-label="divider" />
                         )}
-
                       <ChipsItem
                         chip={chip}
                         chipContainerStyle={styles?.chipContainerStyle}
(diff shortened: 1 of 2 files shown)


components/chips.tsx
```diff
@@ -668,10 +667,10 @@ function ChipsItem({
     <ChipItemWrapper
       $hovered={hovered === chip.id}
       $style={chipContainerStyle}
-      onClick={async (e) => {
-        await e.stopPropagation();
+      onMouseDown={async (e) => {
+        e.preventDefault();
+        e.stopPropagation();
         await onOptionClicked?.(chip);
-        await inputRef.current.focus();
```
adamnoto, line 674: interesting, why does this code fixed the focus issue?

alimnfl: yeah, it because of the `onClick` would broke the focus, before we need to refocus by `inputRef.current.focus()`, that's why the appearance is not consistent

after quite takes time for experiment... some case by this would be easier to solved :)

only using `onMouseDown`, would not broken and get what we want, so even not add re-focus, they will not unfocused....

adamnoto (approved):
OK!

Image

PR #449, opened 2026-06-26
[Improvement] SYST-759: Set default background on the Capsule (not inherit)

alimnfl:
Description:
Previously, The `Capsule` component shows transparent background, which caused inconsistent rendering because it did not provides a default background color.

Additionally, in the the error, the Capsule should display a red border to indicate that the entered value is invalid or doesn't match the expected value.

This ticket addresses these issues by introducing a default background color for the `Capsule` component and applying a red border when the Capsule is in an error state, and related for `styling` in `StatefulForm.mobile`

Before:
image (17)

After:
Screenshot 2026-06-26 at 17 02 24

Source:
[[Improvement] SYST-759: Set default background on the Capsule (not inherit)](https://linear.app/systatum/issue/SYST-759/set-default-background-on-the-capsule-not-inherit)

Tick what you have done:
[x] I have tested it myself


adamnoto (approved):
Image

PR #451, opened 2026-06-29
[Bug] SYST-765: Make defining `id` and `className` works on `Title` rightSection and possibly other sections

alimnfl:
Description:
This pull request fixes an issue where the `id` and `className` props were not applied correctly to the action button.

The root cause was that the `ContextMenu` component was being used in a scenario where additional props needed to be forwarded, which it was not designed to support. To address this, the implementation has been refactored to use our `Button` component directly.

Since the current design does not require switching to a hamburger menu when there are multiple actions, using `Button` is a simpler and more appropriate solution. This refactor also preserves the existing behavior while ensuring that `id`, `className`, and other related props are forwarded and work as expected.

Source:
[[Bug] SYST-765: Make defining `id` and `className` works on `Title` rightSection and possibly other sections](https://linear.app/systatum/issue/SYST-765/make-defining-id-and-classname-works-on-title-rightsection-and)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/context-menu.tsx
+++ b/components/context-menu.tsx
@@ -92,8 +92,9 @@ export default function ContextMenu({
         ? buttonTheme?.[variant]?.hoverBackgroundColor
         : hoverBackgroundColor;
 
-      const resolvedButtonProps = {
+      const resolvedButtonProps: ButtonProps = {
         ...buttonProps,
+        ...action,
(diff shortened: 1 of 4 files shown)


adamnoto (approved):
finnncncnc

PR #453, opened 2026-06-29
[Improvement] SYST-768: Improves `Radio` hover behavior

alimnfl:
Description:
Previously, the `Radio` component had inconsistent hover styling compared to the `Checkbox`. Unlike `Checkbox`, which always uses the theme's base background color for the input element, `Radio` inherited its background color, resulting in an inconsistent and unnatural appearance.

And at the same time we need adjustment on the `StatefulForm` mobile, which really weird when hovering shows the `background` color, we need adjust the `color` for whole `label` inside of `Radio`.

This ticket addresses both issues by:

- Updating the `Radio` component to use the correct base background color instead of inheriting it.
- Adjusting the hover styling so the background is applied across the entire radio label, providing a more consistent experience in `StatefulForm.Mobile`.

All issue and result shows in source below.

Source:
[[Improvement] SYST-768: Improves `Radio` hover behavior](https://linear.app/systatum/issue/SYST-768/improves-radio-hover-behavior)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/checkbox.tsx
+++ b/components/checkbox.tsx
@@ -70,7 +70,7 @@ function BaseCheckbox({
 
   return (
     <InputWrapper
-      aria-label="input-wrapper-checkbox"
+      aria-label="checkbox-label"
       htmlFor={props.disabled ? null : id}
       $hasDescription={!!description}
       $highlight={!!highlightOnChecked}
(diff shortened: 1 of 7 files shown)


adamnoto (approved):
Image

PR #454, opened 2026-06-29
[Improvement] SYST-767: Change API of defining column-level action for `Table`

alimnfl:
Description:
This pull request improves the `Table` API action for column level to put only in the `columns` instead of using `subMenuList` and activate by sortable, which the flexibility not same when we put on this API.

```tsx
export interface TableColumn {
  caption: string;
  styles?: TableColumnStyle;
  width?: string;
  id: string;
  actions?: ((id?: string) => TableColumnAction) | TableColumnAction;
}
```

It also another related should be changes at all, and it also ensures the actions works properly 

Source:
[[Improvement] SYST-767: Change API of defining column-level action for `Table`](https://linear.app/systatum/issue/SYST-767/change-api-of-defining-column-level-action-for-table)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/table.tsx
+++ b/components/table.tsx
@@ -43,9 +43,18 @@ export interface TableColumn {
 }
 
 export interface TableColumnAction
-  extends Omit<ActionButtonProps, "showSubMenuOn" | "caption" | "onClick"> {
+  extends Omit<
+    ActionButtonProps,
+    "showSubMenuOn" | "caption" | "onClick" | "variant"
+  > {
   title?: string;
+  variant?: TableColumnVariant;
 }
+
+export type TableColumnVariant = Exclude<
+  ButtonProps["variant"],
+  `outline-${string}`
+>;
 
 export type TableSubMenuList = TipMenuItemProps;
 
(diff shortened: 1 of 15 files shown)


adamnoto (changes requested):
- No need "With sorter" button on the table to activate the sorter, ALWAYS activate the sorter for "Loose"
- Make hovering background color to `rgb(52, 52, 52)` for dark, for a better visibility (this is for all table)
- Do not change column's height when there's sorter actions (this is for all table). Currently if a table has sorter action, the column height is bigger
  
  Image

  Versus:

  Image

  Making it having sorter on the column must not change its height.

components/table.tsx
```diff
@@ -33,20 +32,35 @@ import { TableThemeConfig } from "./../theme";
...
 import { Scrollbar, ScrollbarRef } from "./scrollbar";
+import { Button, ButtonProps } from "./button";
 
 export interface TableColumn {
   caption: string;
-  sortable?: boolean;
   styles?: TableColumnStyle;
   width?: string;
   id: string;
+  actions?: ((id?: string) => TableColumnAction) | TableColumnAction;
```
adamnoto: please always make it function; I think that's better, ie consistent. (as we agreed from our discussion)

alimnfl:
> * No need "With sorter" button on the table to activate the sorter, ALWAYS activate the sorter for "Loose"
> * Make hovering background color to `rgb(52, 52, 52)` for dark, for a better visibility (this is for all table)
> * Do not change column's height when there's sorter actions (this is for all table). Currently if a table has sorter action, the column height is bigger
>   Image
>   Versus:
>   Image
>   Making it having sorter on the column must not change its height.

done

components/table.tsx
alimnfl: done

adamnoto (approved):
Image

PR #455, opened 2026-06-30
[Bug] SYST-769: Fix cannot build on `main`

alimnfl:
Description:
This pull request fixes an issue when we build the `component` inside of the coneto, it would be error, caused from `mobile` prop in the `Title` was deleted.

Source:
[[Bug] SYST-769: Fix cannot build on `main`](https://linear.app/systatum/issue/SYST-769/fix-cannot-build-on-main)

Tick what you have done:
[x] I have tested it myself


adamnoto (approved):
Image

PR #456, opened 2026-06-30
[Mobile] SYST-762: Add `mobile` to `Datebox`

alimnfl:
Description:
This change introduces a `mobile` prop to the `Datebox` component so it can render differently on mobile devices.

Instead of rendering the calendar popup attached to the input element, the `Datebox` should renders it at the bottom of the screen. The appearance would be similar like `Combobox` component.

Source:
[[Mobile] SYST-762: Add `mobile` to `Datebox`](https://linear.app/systatum/issue/SYST-762/add-mobile-to-datebox)

Tick what you have done:
[x] I have tested it myself


adamnoto (approved):
Image

PR #462, opened 2026-07-02
[Bug] SYST-774: Refine `Chips` badge `appearance` and `interaction`

alimnfl:
Description:
This ticket fixes an issue with **Chips** badge coloring. When a badge provided a `backgroundColor`, the option in the dropdown incorrectly inherited that background color. After introducing `transparent` as the default appearance, the option background always remained transparent, even when a custom color was provided.

**Source issue:**

* https://coneto.systatum.com/index.html?path=/story/input-elements-chips--dark-background

After this change, the **Dark Background** story renders as expected.

Additionally, this ticket fixes badge action behavior. Clicking a badge action should no longer trigger option selection, allowing the action to behave independently.

**Source issue:**

* https://coneto.systatum.com/index.html?path=/story/input-elements-chips--custom-renderer

To achieve this:

* Badge actions now use `onMouseDown` instead of `onClick` to prevent the input from losing focus when the action is clicked.
* The badge action callback has been enhanced to expose both the badge and the mouse event, allowing consumers to implement custom behavior.

```tsx
export interface BadgeAction extends Omit<BaseAction, "onClick" | "caption"> {
  onClick?: (props: { badge?: BadgeProps; event?: React.MouseEvent }) => void;
  size?: number;
  styles?: ButtonStyles;
  title?: string;
}
```

Finally, this ticket also removes the incorrect styling in the **Custom Renderer** [story](https://coneto.systatum.com/index.html?path=/story/input-elements-chips--custom-renderer).

Source:
[[Bug] SYST-774: Refine `Chips` badge `appearance` and `interaction`](https://linear.app/systatum/issue/SYST-774/refine-chips-badge-appearance-and-interaction)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/badge.tsx
+++ b/components/badge.tsx
@@ -37,7 +37,7 @@ export interface BadgeStyles {
 }
 
 export interface BadgeAction extends Omit<BaseAction, "onClick" | "caption"> {
-  onClick?: (badge?: BadgeProps) => void;
+  onClick?: (props: { badge?: BadgeProps; event?: React.MouseEvent }) => void;
   size?: number;
   styles?: ButtonStyles;
   title?: string;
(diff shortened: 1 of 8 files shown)


adamnoto (changes requested):
Extreme bug: we can't choose [chips on stateful form](https://dev.files.systatum.com/coneto/e4abace0/index.html?path=/story/input-elements-statefulform--all-case) in macOS FireFox, can you please test.

Maybe let's test against chrome and firefox automatically?

alimnfl:
> Extreme bug: we can't choose [chips on stateful form](https://dev.files.systatum.com/coneto/e4abace0/index.html?path=/story/input-elements-statefulform--all-case) in macOS FireFox, can you please test.
> 
> Maybe let's test against chrome and firefox automatically?

was done for this, it was changed actually in `Stateful` PR for logical, but we still only make from main, now all good I think for `chips` case.

adamnoto (approved):
Image

PR #463, opened 2026-07-03
[Bug] SYST-775: Calendar stretch after open the `picker` mode

alimnfl:
Description:
This pull request fixes an issue where the `Calendar` component stretched after opening the changes mode for the `month` or `year` picker. 

The issue was caused by the `Calendar` wrapper not preserving a consistent width. This change adds the necessary guard to keep the Calendar width stable and ensures the issue does not occur again.

Additionally, this pull request fixes the height issue, now we're using `height` instead of `min-width` in the drawer. This combination caused the drawer to incorrectly expand and render for some case.

Source:
[[Bug] SYST-775: Calendar stretch after open the `picker` mode](https://linear.app/systatum/issue/SYST-775/calendar-stretch-after-open-the-picker-mode)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/calendar.tsx
+++ b/components/calendar.tsx
@@ -560,6 +560,7 @@ function BaseCalendar({
               padding: 0.5rem;
               font-size: 0.875rem;
               min-width: 300px;
+              max-width: 300px;
               display: flex;
               flex-direction: column;
               gap: 0.5rem;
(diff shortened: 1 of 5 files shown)


adamnoto (approved):
Image

PR #469, opened 2026-07-07
[Mobile] SYST-782: Make split pane separator adjustable even using tap

alimnfl:
Description:
This pull request improves the function to use `pointer` instead of using the `mouse`, it would not support in mobile behavior when dragging or resizing the separation content. It also ensures the pointer behavior would works properly in test, either by cypress or by our phone.

Source:
[[Mobile] SYST-782: Make split pane separator adjustable even using tap](https://linear.app/systatum/issue/SYST-782/make-split-pane-separator-adjustable-even-using-tap)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/split-pane.tsx
+++ b/components/split-pane.tsx
@@ -138,7 +138,7 @@ function SplitPane({
     [isVertical]
   );
 
-  const stopDrag = useCallback(
+  const onPointerUp = useCallback(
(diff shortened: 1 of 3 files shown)


adamnoto (approved):
Image

PR #470, opened 2026-07-07
[Bug] SYST-786: `PaperDialog` issue on `Android`

alimnfl:
Description:
This pull request fixes the behavior on `Android` , after catch an issue in the `bil-quran` app, the `PaperDialog` doesn't work properly for `resizeable`, we should provides the `touch-action: none` on the drag, the real reason it causes from between `native` android gesture and `javascript` scroll, so we need to added it.

It also ensures the drag indicator provides properly, and make sure in real project, or in another devices would works correctly.

Source:
[[Bug] SYST-786: `PaperDialog` issue on `Android`](https://linear.app/systatum/issue/SYST-786/paperdialog-issue-on-android)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/paper-dialog.tsx
+++ b/components/paper-dialog.tsx
@@ -810,6 +810,7 @@ const MotionDialog = styled(motion.div)<{
 `;
 
 const DesktopResizeHandle = styled.div<{ $isLeft: boolean }>`
+  touch-action: none;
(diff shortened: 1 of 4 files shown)


adamnoto (approved):
Image

PR #473, opened 2026-07-09
[Bug] SYST-791: Refactors `width` uses if not using `mobile` on the `StatefulForm.Combobox`

alimnfl:
Description:
Previously, The `combobox` inside of the `Stateful` causes bug for uses `min-width: 140px` in `Combobox` after `introduces` a `mobile` appearance. This pull request fixes that issue. It also ensures the `width` works properly if using `percentage` in `non-mobile` case, if not using mobile appearance.

Screenshot 2026-07-09 at 11 46 13

Source:
[[Bug] SYST-791: Refactors `width` uses if not using `mobile` on the `StatefulForm.Combobox`](https://linear.app/systatum/issue/SYST-791/refactors-width-uses-if-not-using-mobile-on-the-statefulformcombobox)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/stateful-form.tsx
+++ b/components/stateful-form.tsx
@@ -2200,7 +2200,8 @@ function FormFields<T extends FieldValues>({
                                 ${field.combobox?.styles?.bodyStyle}
                               `,
                               controlStyle: css`
-                                ${field.title &&
+                                ${mobile &&
+                                field.title &&
                                 css`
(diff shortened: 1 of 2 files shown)


adamnoto (approved):
Image

Ok, please try giving before-after next time for easier/quicker review :)

PR #474, opened 2026-07-09
[Bug] SYST-792: Fix ugly loose table when it has only a few data

alimnfl:
Description:
This pull request fixes the `table` component if only add a few column in condition provides `loose` as a feature. Previously causes the bug because the `flex` it was unset, so it's not make the width always full, after introduces the `loose` not use some content less than 5 column or more.

It also ensures the `loose` feature still works properly even we given a few column inside of the `Table`. 

Before:
image

After:
image

Source:
[[Bug] SYST-792: Fix ugly loose table when it has only a few data](https://linear.app/systatum/issue/SYST-792/fix-ugly-loose-table-when-it-has-only-a-few-data)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/scrollbar.tsx
+++ b/components/scrollbar.tsx
@@ -17,6 +17,8 @@ interface ScrollbarProps {
   autoHideDelay?: number;
   onScroll?: () => void;
   style?: CSSProperties;
+  totalSize?: number;
+  scrollOffset?: number;
 }
 
 export interface ScrollbarRef {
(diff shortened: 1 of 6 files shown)


adamnoto (changes requested):
Weird story rendering

Image

alimnfl:
> Weird story rendering
> 
> Image

done

adamnoto (approved):
Image

PR #475, opened 2026-07-10
[Bug] SYST-793: Fix `virtualization` on the `Table`

alimnfl:
Description:
This pull request fixes the `virtualization` issue when not get the real scroll element after fetch the `content`. It also ensures not shows the content on the table never  happen again.

Source:
[[Bug] SYST-793: Fix `virtualization` on the `Table`](https://linear.app/systatum/issue/SYST-793/fix-virtualization-on-the-table)

Tick what you have done:
[x] I have tested it myself


adamnoto (approved):
Image

PR #478, opened 2026-07-14
[Bug] SYST-800: `NavTab` tab style issue

alimnfl:
Description:
Previously, the `NavTab` would be `error` if we use the `tabs` less than 4 item, the `width` would be 100%, but not the item, so the content shows not correctly, especially in the `desktop` appearance, and the `tabStyle` now not added properly, which causes an issue like on below

Screenshot 2026-07-14 at 10 52 14

And if not given the `title`, the `tab` it would be broken, because the `align-items` using `end`, so we should use by `stretch`, 

Screenshot 2026-07-14 at 11 24 55

This pull request fixes several issues in the `NavTab` component and includes the following changes:
1. Refactored the placement of `tabStyle` by moving it out of `styles` for better organization.
2. Improved the width calculation for tabs on mobile. The same logic is now applied to the `Tooltip` to keep both components consistent:
   * Fewer than 4 tabs: each tab has an equal width.
   * 4 or more tabs: each tab uses `width: 100%`.
3. Updated the height layout by using `height: auto` and `height: 100%` where appropriate, and added `align-items: stretch` to ensure tabs without a `title` still have a clean and consistent appearance.
4. Set `justify-content: space-between` as the default for the wrapper while still allowing users to override it (for example, with `justify-content: center`) through custom styles.

Screenshot 2026-07-14 at 11 06 59

Source:
[[Bug] SYST-800: `NavTab` tab style issue](https://linear.app/systatum/issue/SYST-800/navtab-tab-style-issue)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/nav-tab.tsx
+++ b/components/nav-tab.tsx
@@ -336,16 +336,21 @@ function NavTab({
                   containerStyle: css`
                     width: fit-content;
                     ${mobile &&
-                    css`
-                      width: 100%;
-                    `};
+                    getTabWidth({
+                      mobile,
+                      tabsLength: visibleTabs?.length,
+                      width: maxTabWidth,
+                    })};
+
                     ${styles?.tabStyle}
                   `,
                   triggerStyle: css`
                     ${mobile &&
-                    css`
-                      width: 100%;
-                    `};
+                    getTabWidth({
+                      mobile,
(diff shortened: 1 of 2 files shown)


adamnoto (approved):
Image

PR #481, opened 2026-07-15
[Bug] SYST-801: Fix `Combobox.drawer` bug in Coneto

alimnfl:
Description:
This pull request fixes an issue on the `Combobox` component if `options` where providing 5 `options` resulted in an incorrect appearance with extra empty space. This issue was caused by the height calculation logic, which only handled cases where the number of options was `less than 5`

The fix updates the condition to `use less than or equal 5`, instead of just `less than 5`.

```tsx
const hasFewOptions = finalOptions?.length <= 5
```

This ensures the `Combobox` uses the correct height for 5 or fewer options, preventing the extra empty space and avoiding the same issue in the future.

Source:
[[Bug] SYST-801: Fix dropdown bug in Coneto](https://linear.app/systatum/issue/SYST-801/fix-dropdown-bug-in-coneto)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/combobox.tsx
+++ b/components/combobox.tsx
@@ -877,7 +877,7 @@ function ComboboxDrawer({
 
   const searchboxProps = typeof searchbox === "object" && searchbox;
 
-  const hasFewOptions = finalOptions?.length < 5;
+  const hasFewOptions = finalOptions?.length <= 5;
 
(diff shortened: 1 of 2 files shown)


adamnoto (approved):
Image

PR #485, opened 2026-07-20
[Mobile] SYST-808: Add `width`, `height`, and changes the `closable` behavior in `ScreenTransition`

alimnfl:
Description:
This pull request enhances the `ScreenTransition` component by allowing each screen to define its own initial `width`, `height`, and `closable` behavior.

```ts
type ScreenConfig = {
  component: ScreensComponent;
  sheet?: ScreenSheetConfig;
  width?: string;
  height?: string;
  closable?: boolean;
};
```

It also introduces customizable `indicatorStyle`, `contentStyle`, and `containerStyle` properties, providing greater flexibility when styling the underlying `PaperDialog`. These style props allow consumers to customize the dialog directly without introducing additional wrapper elements.

```ts
export interface ScreenTransitionProps<TScreens extends ScreensMap> {
  /** Registry of every screen this transition can render, keyed by id */
  screens: TScreens;
  /** Ordered stack of active screen keys. First = base, rest = nested dialogs */
  activeScreens: (keyof TScreens)[] | string[];
  /** Called with the next stack whenever navigation happens */
  onScreenChange: (screens: (keyof TScreens)[]) => void;
  /** Styles applied to the underlying PaperDialog */
  styles?: ScreenTransitionStyles;
}

export type ScreenTransitionStyles = Pick<
  PaperDialogStyles,
  "indicatorStyle" | "contentStyle" | "containerStyle"
>;
```

When a screen is presented as a mobile sheet, the Escape key is no longer enabled by default. Since sheet presentation is intended for mobile interactions, supporting desktop-specific dismissal via the Escape key is unnecessary. The `withEscape` option remains available and continues to default to `true` when using `PaperDialog` directly.

```ts
export interface PaperDialogClosable {
  withOverlay?: boolean;
  withEscape?: boolean;
  withButton?: boolean;
  withIndicator?: boolean;
}
```

The `PaperDialog` `onChange` callback has also been enhanced to include a trigger parameter that identifies what caused the state transition. This allows consumers to distinguish between programmatic changes and user interactions such as clicking the overlay, pressing the Escape key, dragging, resizing, or interacting with dialog controls.

```ts
export const PaperDialogTrigger = {
  API: "api",
  Overlay: "overlay",
  Escape: "escape",
  Drag: "drag",
  Control: "control",
} as const;

export type PaperDialogTrigger =
  (typeof PaperDialogTrigger)[keyof typeof PaperDialogTrigger];

onChange?: (
  state: PaperDialogState,
  trigger: PaperDialogTrigger
) => void;
```

Consumers can use this trigger to determine what initiated a dialog state change, enabling more granular control over dialog behavior.

Source:
[[Mobile] SYST-808: Add `width`, `height`, and changes the `closable` behavior in `ScreenTransition`](https://linear.app/systatum/issue/SYST-808/add-width-height-and-changes-the-closable-behavior-in-screentransition)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/paper-dialog.tsx
+++ b/components/paper-dialog.tsx
@@ -445,7 +445,7 @@ const PaperDialog = forwardRef<PaperDialogRef, PaperDialogProps>(
     );
 
     const closeDialog = useCallback(
-      async (withTimeout: boolean = true, trigger: PaperDialogTrigger) => {
+      async (withTimeout: boolean, trigger: PaperDialogTrigger) => {
         const close = async () =>
           await handleChangeDialog(PaperDialogState.Closed, trigger);
 
@@ -491,7 +491,7 @@ const PaperDialog = forwardRef<PaperDialogRef, PaperDialogProps>(
           canCloseWithEscape &&
           (dialogState === "restored" || dialogState === "minimized")
         ) {
-          closeDialog(true, PaperDialogTrigger.Escape);
+          closeDialog(false, PaperDialogTrigger.Escape);
           setShowTitlebar(false);
         }
       },
--- a/components/paper-dialog.tsx
+++ b/components/paper-dialog.tsx
@@ -33,7 +33,7 @@ export type PaperDialogState =
   (typeof PaperDialogState)[keyof typeof PaperDialogState];
 
 export const PaperDialogTrigger = {
-  Api: "api",
+  API: "api",
   Overlay: "overlay",
   Escape: "escape",
   Drag: "drag",
@@ -78,6 +78,7 @@ export interface PaperDialogClosable {
   withOverlay?: boolean;
   withEscape?: boolean;
   withButton?: boolean;
+  withIndicator?: boolean;
 }
 
 export interface PaperDialogResizable {
@@ -122,8 +123,12 @@ export interface PaperDialogContentStyles {
 
 export interface PaperDialogRef {
   openDialog: () => void;
-  closeDialog: (withTimeout?: boolean) => void;
+  closeDialog: (props?: PaperDialogCloseOption) => void;
   minimizeDialog: () => void;
+}
+export interface PaperDialogCloseOption {
+  withMinimize?: boolean;
+  withTimeout?: boolean;
 }
 
 const PaperDialog = forwardRef<PaperDialogRef, PaperDialogProps>(
@@ -164,6 +169,7 @@ const PaperDialog = forwardRef<PaperDialogRef, PaperDialogProps>(
       withEscape = true,
       withOverlay = true,
       withButton = true,
+      withIndicator = true,
     } = typeof closable === "object" ? closable : {};
 
     const canCloseWithEscape =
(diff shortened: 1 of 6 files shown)


components/paper-dialog.tsx
adamnoto: still treats closable as a plain boolean, an object like `{ withOverlay: false, withEscape: false, withButton: false }` is truthy, let's make it more accurate

components/screen-transition.tsx
```diff
@@ -92,7 +106,7 @@ function ScreenTransition<TScreens extends ScreensMap>({
   const closingIndicesRef = useRef<Set<number>>(new Set());
```
adamnoto: is it used?

components/paper-dialog.tsx
```diff
@@ -395,24 +435,25 @@ const PaperDialog = forwardRef<PaperDialogRef, PaperDialogProps>(
...
 
         await setResizeHeight(null);
         await setResizeWidth(null);
 
-        if (mobile) {
-          await handleChangeDialog("minimized");
+        if (withTimeout) {
+          await handleChangeDialog(PaperDialogState.Minimized, trigger);
 
           if (withTimeout) {
```
adamnoto, line 479: double if withTimeout? please simplify

components/paper-dialog.tsx
```diff
@@ -32,6 +32,18 @@ export const PaperDialogState = {
 export type PaperDialogState =
   (typeof PaperDialogState)[keyof typeof PaperDialogState];
 
+export const PaperDialogTrigger = {
+  Api: "api",
```
adamnoto: `API`

components/paper-dialog.tsx
```diff
@@ -32,6 +32,18 @@ export const PaperDialogState = {
...
+  Overlay: "overlay",
+  Escape: "escape",
+  Drag: "drag",
+  Resize: "resize",
+  Control: "control",
```
adamnoto, line 49: how is `drag` and `resize` different from `control` ? I thought control is for manual button mechanism right?

adamnoto (changes requested):
ALSO `closable: false` on a sheet screen doesn't actually block dismissal (highest severity). So a screen configured with `{ sheet: true, closable: false }` can still be swiped away. Please fix and add test.

components/paper-dialog.tsx
alimnfl: done

components/paper-dialog.tsx
alimnfl: yes control for button

components/screen-transition.tsx
alimnfl: yes, isn't

components/paper-dialog.tsx
alimnfl: done

components/paper-dialog.tsx
alimnfl: just for note, differentiate `drag` and `resize` is for the trigger state between with or without `resizable` mode, 

so do you want to make those only in one option?

adamnoto (approved):
Image

PR #486, opened 2026-07-21
[Bug] SYST-809: Fix hidden condition on `Table.Loose`

alimnfl:
Description:
This ticket fixes the `Table.Loose` header condition when row actions are marked with `hidden: true`, which was previously not handled correctly. It also ensures the test to prevents the `loose.effect` with `hidden: true` not being applied incorrectly again.

Source:
[[Bug] SYST-809: Fix hidden condition on`Table.Loose`](https://linear.app/systatum/issue/SYST-809/fix-hidden-condition-ontableloose)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/table.tsx
+++ b/components/table.tsx
@@ -1539,7 +1539,7 @@ function TableRow({
 
   const { loose, setWithRowActions, isScrolledRight } = useTableLoose();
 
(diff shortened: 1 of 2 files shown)


adamnoto (approved):
Image

PR #488, opened 2026-07-24
[Improvement] SYST-811: Fix `scrollbar`, removes the laggy on the `Table`

alimnfl:
Description:
This ticket improves all behavior on the `Table` component, which fixing the scroll, the rendering, because previously indexing always added after scrolling, since we use by `virtualize`, only shows data with `20` and shows, so this is step by step how to fix it:

1. Removes `resizeObserver` on the scrollbar, which we can always adapt on the scroll `thumb`
2. Removes uses of `cloneElement` to the all element, even for `Table` parent level, `row` level, `row.group` level, and `row.cell`, so how to get data? using `context` to give another element same value
3. Add memoization to prevent unnecessary re-renders in certain cases. `React.memo` is now applied at the `Row.Cell` level to better preserve the component architecture.

Additionally I catch some bug inside on the `group` level if scroll in that element, not really stick after div element uses on the `position: absolute` in div element, so still need to use top with depends on the `virtualRow.start`

broke screenshot:

Screenshot 2026-07-24 at 15 25 17

from:

```
 <div
  key={virtualRow.key}
  data-index={virtualRow.index}
  ref={rowVirtualizer.measureElement}
  style={{
    position: "absolute",
    left: 0,
    width: "100%",
    display: "flex",
    flexDirection: "column",
    transform: translateY....
  }}
>
 {children}
</div>
```

to:

```
 <div
  key={virtualRow.key}
  data-index={virtualRow.index}
  ref={rowVirtualizer.measureElement}
  style={{
    position: "absolute",
    top: virtualRow.start,
    left: 0,
    width: "100%",
    display: "flex",
    flexDirection: "column",
  }}
>
 {children}
</div>
```

Source:
[[Improvement] SYST-811: Fix `scrollbar`, removes the laggy on the `Table`](https://linear.app/systatum/issue/SYST-811/fix-scrollbar-removes-the-laggy-on-the-table)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/scrollbar.tsx
+++ b/components/scrollbar.tsx
@@ -128,27 +128,26 @@ const Scrollbar = forwardRef<ScrollbarRef, ScrollbarProps>(
     // totalSize/scrollOffset props to propagate through a re-render.
     const updateControlledThumbY = useCallback(() => {
       const el = viewportRef.current;
-      if (!el || !isControlledY) return;
+      if (!el) return;
 
-      const { totalSize, scrollOffset } = controlledRef.current;
-      const clientHeight = el.clientHeight;
+      const { scrollTop, clientHeight } = el; // <- read live, not from props
+      const effectiveScrollHeight =
+        controlledRef.current.totalSize ?? el.scrollHeight;
 
-      if (!clientHeight || (totalSize ?? 0) <= clientHeight) {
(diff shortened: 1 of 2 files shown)


adamnoto (approved):
Image

PR #489, opened 2026-07-27
[Bug] SYST-739: Do not render title right section if there's no right section

alimnfl:
Description:
This pull request fixes an issue where the `rightSection` was rendered even when no `headerActions` were provided, caused by incorrect logic when handling the actions array.

Additionally, it introduces support for an `icon` in the `Card` component using the `Figure` component, and makes the card content optional. This provides greater flexibility, allowing consumers to render only a `Title`, `Pretitle`, or even just an `icon` without requiring card content.

It also ensures that the `rightSection` is not rendered unless header actions are explicitly provided.

Source:
[[Bug] SYST-739: Do not render title right section if there's no right section](https://linear.app/systatum/issue/SYST-739/do-not-render-title-right-section-if-theres-no-right-section)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/card.stories.tsx
+++ b/components/card.stories.tsx
@@ -1122,7 +1122,7 @@ export const Toggleable: Story = {
 
       const sortableColumns: TableColumn[] = columns.map((column) => ({
         ...column,
-        actions: {
+        actions: () => ({
           title: "Sort",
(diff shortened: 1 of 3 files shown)


adamnoto (approved):
Image

PR #490, opened 2026-07-28
Fix sluggish table horizontal scrolling

adamnoto:
## Root causes & fixes

1. Header/body/summary scroll sync echoed through all three containers every tick, redoing scrollWidth/clientWidth reads (forcing sticky-layout recalcs) 3x per scroll. Added a re-entrancy guard so echoed events short-circuit.

2. The scroll-shadow flags lived in React context, so crossing either edge re-rendered every cell in every row/column. Replaced with CSS custom properties written directly on the container ref.

Tick what you have done:

[ ] I have double checked the functionality with the ticket in linear, or any other relevant discussion avenue
[x] I have tested it myself

### DIFF
--- a/components/table.tsx
+++ b/components/table.tsx
@@ -1,6 +1,5 @@
 import React, {
   Children,
-  cloneElement,
   createContext,
   isValidElement,
   ReactElement,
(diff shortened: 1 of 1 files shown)


components/table.tsx
```diff
@@ -724,11 +780,12 @@ function Table({
                                   ref={rowVirtualizer.measureElement}
                                   style={{
                                     position: "absolute",
-                                    top: virtualRow.start,
+                                    top: 0,
                                     left: 0,
                                     width: "100%",
                                     display: "flex",
                                     flexDirection: "column",
+                                    transform: `translateY(${virtualRow.start}px)`,
```
alimnfl: please not using `translateY` for our table `virtualizer`, since we use sticky on the `Table.Group`, if using `top: 0`, the sticky behavior doesn't know the top is it, 

because the container is "absolute"

Image

PR #494, opened 2026-07-31
[Improvement] SYST-816: Add a `scoreLabel` prop for flexibility in `Rating`

alimnfl:
Description:
This pull request replaces the `withLabel` prop with a `scoreLabel` prop to provide more flexibility when rendering the Rating label. Previously, `withLabel` only displayed the value in the `${value} / ${maxValue}` format, which was limiting.

```tsx
export type RatingScoreLabelRender =
  ReactNode | ((props?: { value?: number; maxValue?: number }) => ReactNode);

export interface RatingScoreLabel {
  text?: RatingScoreLabelRender;
  position?: RatingScoreLabelPosition;
}
```

It also changes the `Rating` component so that it no longer relies on the `editable` prop to determine whether the value can be changed. Instead, the Rating becomes editable whenever an `onChange` handler is provided.

Additionally, this ticket updates the component stories to better demonstrate the intended usage of the Rating component.

Source:
[[Improvement] SYST-816: Add a `renderLabel` prop for flexibility in `Rating`](https://linear.app/systatum/issue/SYST-816/add-a-iconSidelabel-prop-for-flexibility-in-rating)

Tick what you have done:
[x] I have tested it myself


adamnoto (changes requested):
let's generalize to `label` instead then, if not given (default) assume behavior like withLabel = false

alimnfl:
> let's generalize to `label` instead then, if not given (default) assume behavior like withLabel = false

we can't, since this related to `FieldLane` component, this is example how to use `label` prop now:

image

any suggestion?

adamnoto (changes requested):
ok `iconSideLabel` then?

alimnfl:
> ok `iconSideLabel` then?

now using `iconSideLabel`

adamnoto (changes requested):
so sorry Alim, can we do:

`scoreLabel={text: /* your change here, I guess function right? */, position: "right" or "left"}`

so the prop is `scoreLabel` and it's expecting an object 🙏

alimnfl:
> so sorry Alim, can we do:
> 
> `scoreLabel={text: /* your change here, I guess function right? */, position: "right" or "left"}`
> 
> so the prop is `scoreLabel` and it's expecting an object 🙏

done

adamnoto (approved):
Image

PR #495, opened 2026-07-31
Separate line between title into 2 line components

adamnoto:
This way, we don't need to have background on the label anymore.

Tick what you have done:

[X] I have double checked the functionality with the ticket in linear, or any other relevant discussion avenue
[X] I have created/updated any relevant test code
[X] I have tested it myself


alimnfl (approved):
I think all good for now, so we were not used `absolute` for this case, and separate by 2 line, okay then....

yay-kitty

PR #497, opened 2026-08-03
Make table overlay blocker be relative to the table

adamnoto:
Tick what you have done:

[X] I have double checked the functionality with the ticket in linear, or any other relevant discussion avenue
[X] I have created/updated any relevant test code
[X] I have tested it myself


components/overlay-blocker.tsx
```diff
@@ -76,46 +87,62 @@ export const OverlayBlocker = forwardRef<
...
       const body = document.body;
 
-      const prev = {
-        overflow: body.style.overflow,
-        position: body.style.position,
-        top: body.style.top,
-        width: body.style.width,
-      };
+      // `relative` scopes this to its own container, so the rest of the
+      // page must keep scrolling normally - only lock/restore the body
```
alimnfl, line 91: I think this one would be good if we just using by guard negative from `relative` prop, here, instead of using ternary

```tsx
 useEffect(() => {
      if (!visible) return;

      if (document.querySelector(".sbdocs-content")) return;

      const safeRegions = exemptRegions ?? [];

      const scrollY = window.scrollY;
      const body = document.body;

      // `relative` scopes this to its own container, so the rest of the
      // page must keep scrolling normally - only lock/restore the body
      // when blocking the whole viewport.
      const prev = {
        overflow: body.style.overflow,
        position: body.style.position,
        top: body.style.top,
        width: body.style.width,
      };

      if (!relative) {
        body.style.overflow = "hidden";
        body.style.position = "fixed";
        body.style.top = `-${scrollY}px`;
        body.style.width = "100%";
      }

      const allow = (target: EventTarget | null) =>
        isInSafeZone(target, safeRegions);

      // When relative, only wheel/touch events landing on this overlay's
      // own footprint are blocked - everything else on the page scrolls
      // through untouched.
      const withinScope = (target: EventTarget | null) =>
        !relative ||
        (target instanceof Node && overlayNodeRef.current?.contains(target));

      const blockWheel = (e: WheelEvent) => {
        if (allow(e.target) || !withinScope(e.target)) return;
        e.preventDefault();
      };

      const blockTouch = (e: TouchEvent) => {
        if (allow(e.target) || !withinScope(e.target)) return;
        e.preventDefault();
      };

      window.addEventListener("wheel", blockWheel, { passive: false });
      window.addEventListener("touchmove", blockTouch, { passive: false });

      return () => {
        if (!relative) {
          body.style.overflow = prev.overflow;
          body.style.position = prev.position;
          body.style.top = prev.top;
          body.style.width = prev.width;

          window.scrollTo(0, scrollY);
        }

        window.removeEventListener("wheel", blockWheel);
        window.removeEventListener("touchmove", blockTouch);
      };
    }, [exemptRegions, visible, relative]);
```

PR #498, opened 2026-08-04
[Improvement] SYST-819: Extends the some `HTMLDivElement` inside of `ThumbField`

alimnfl:
Description:
This ticket adds a `HTMLDivElement` props to `ThumbField`. Would it be better to expose them at the `InputGroup` level for more flexibility handle events, such as `onClick`, `onMouseEnter`, and others.

Source:
[[Improvement] SYST-819: Extends the `HTMLDivElement` inside of `ThumbField`](https://linear.app/systatum/issue/SYST-819/extends-the-htmldivelement-inside-of-thumbfield)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/thumb-field.tsx
+++ b/components/thumb-field.tsx
@@ -25,6 +25,10 @@ interface BaseThumbFieldProps
   id?: string;
   showError?: boolean;
   thumbText?: ThumbFieldThumbText;
+  onClick?: (event: React.MouseEvent<HTMLDivElement>) => void;
+  onMouseDown?: (event: React.MouseEvent<HTMLDivElement>) => void;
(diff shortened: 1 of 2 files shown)


adamnoto (changes requested):
can we just expose as necessary, so if you need onClick, onMouseDown, etc, can we just expose them one by one?

alimnfl:
> can we just expose as necessary, so if you need onClick, onMouseDown, etc, can we just expose them one by one?

sure... done for it

adamnoto (approved):
catyay

I don't like the culture of over exposing HTML attributes; let's just ALWAYS ALWAYS please ALWAYS: add as we need it

PR #499, opened 2026-08-05
[Research] SYST-820: Use `RHF` with `values`, instead of `defaultValues`

alimnfl:
Description:
This pull request migrates `StatefulForm` from React Hook Form's (`RHF`) `defaultValues` to the reactive `values` API, enabling fully controlled form state and reactive synchronization with external updates.

  | defaultValues | values
-- | -- | --
When it's applied | Once, on mount only | Reactively, every time the object reference changes
Reacts to later prop changes? | ❌ No — ignored after initial mount | ✅ Yes — triggers RHF's internal reset()
Best suited for | Static initial state (e.g. data fetched once before the form renders) | Live external state that needs to stay in sync (e.g. a controlled formValues prop)
Updates raw field values on change? | doesn't update after mount | ✅ Yes
Auto-marks changed fields as touched? | ❌ No | ❌ No
Auto re-runs validation for display (errors)? | ❌ No | ❌ No
Risk if used naively | Form silently stops reflecting new external data | Field values update, but touchedFields/errors silently fall out of sync

### How is the `asynchronous` issue solved?
1. Diffs the latest formValues against prevFormValuesRef.
2. Detects whether the change originated from:
   - Internal user interaction (handleFieldChange via RHF), or
   - External updates (API responses, parent state changes, resets, etc.).
3. Calls:
```tsx
setValue(name, value, {
  shouldValidate: true,
  shouldTouch: true,
  shouldDirty: true,
})
``` 
on genuinely external changes only (via `internallyChangedFieldsRef` to skip changes that already came from `handleFieldChange`)

Source:
[[Research] SYST-820: Use `RHF` with `values`, instead of `defaultValues`](https://linear.app/systatum/issue/SYST-820/use-rhf-with-values-instead-of-defaultvalues)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,6 +1,10 @@
 {
-  "semi": true,
+  "$schema": "./node_modules/oxfmt/configuration_schema.json",
+  "ignorePatterns": [],
+  "semi": false,
   "singleQuote": false,
-  "trailingComma": "es5",
-  "tabWidth": 2
+  "trailingComma": "all",
+  "printWidth": 80,
+  "tabWidth": 2,
+  "endOfLine": "lf"
 }
--- a/components/stateful-form.stories.tsx
+++ b/components/stateful-form.stories.tsx
@@ -1,25 +1,25 @@
-import { Meta, StoryObj } from "@storybook/react/";
-import { useMemo, useState } from "react";
-import styled, { css } from "styled-components";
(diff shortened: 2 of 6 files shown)


adamnoto (approved):
finnxjljxcv

I trust ✌️

PR #500, opened 2026-08-06
[Bug] SYST-821: Fix `LoadingSkeleton` issue if wrapped by another element.

alimnfl:
Description:
Previously, the `LoadingSkeleton` was error if wrapped by some `div` or another `element`, causes the `style` from `parent` level providing by `cloneElement`, for some case is not handled properly, 

Then, after learn case from `Table`, `cloneElement` causing re-renders. This ticket fixes `LoadingSkeleton` with `useContext` to can always get state from parent if `provided`.

Error:
Screenshot 2026-08-06 at 11 12 21

Source:
[[Bug] SYST-821: Fix `LoadingSkeleton` issue when wrapping by another element.](https://linear.app/systatum/issue/SYST-821/fix-loadingskeleton-issue-when-wrapping-by-another-element)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/loading-skeleton.tsx
+++ b/components/loading-skeleton.tsx
@@ -1,10 +1,9 @@
 import {
-  Children,
-  cloneElement,
+  createContext,
   HTMLAttributes,
-  isValidElement,
-  ReactElement,
   ReactNode,
+  useContext,
+  useMemo,
 } from "react";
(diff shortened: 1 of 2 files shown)


adamnoto (approved):
finnsleep

PR #504, opened 2026-08-10
[Bug] SYST-798: Fix rendering of tooltip when not enough space

alimnfl:
Description:
This pull request fixes the `Tooltip` rendering on small screens, especially when the content contains long sentences. Previously, the tooltip always used `white-space: nowrap`, which caused the content to overflow. It now wraps normally.

Additionally, this fixes the `arrow` positioning so it always stays attached to the trigger element. This is done by using `useRef` to reference the trigger element and passing it through `Floating UI` middleware.

Result:
image

Source:
[[Bug] SYST-798: Fix rendering of tooltip when not enough space](https://linear.app/systatum/issue/SYST-798/fix-rendering-of-tooltip-when-not-enough-space)

Tick what you have done:
[x] I have tested it myself

### DIFF
--- a/components/tooltip.tsx
+++ b/components/tooltip.tsx
@@ -3,10 +3,12 @@ import {
   getFloatingPlacement,
 } from "./../lib/floating-placement";
 import {
+  arrow,
   autoUpdate,
   flip,
   offset,
   Placement,
+  shift,
   size,
   useFloating,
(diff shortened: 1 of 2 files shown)


adamnoto (approved):
finnsdjfslf

